PR #32, opened 2014-12-30
Gather index info at the same time as table info, rather than later.

adunstan:
This helps avoid possible problems with later strong table locks.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -143,6 +143,24 @@ const char *PROGRAM_VERSION = "unknown";
 /* Will be used as a unique prefix for advisory locks. */
 #define REPACK_LOCK_PREFIX_STR "16185446"
 
+typedef enum
+{
+	UNPROCESSED,
+	INPROGRESS,
+	FINISHED
+} index_status_t;
+
+/*
+ * per-index information
+ */
+typedef struct repack_index
+{
+	Oid				target_oid;		/* target: OID */
+	const char	   *create_index;	/* CREATE INDEX */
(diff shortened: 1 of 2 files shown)


schmiddy:
Thanks for digging into this, and I see your [description of the problem](http://adpgtech.blogspot.com/2014/12/pgrepack-redux.html). Do you have a suggestion how to reproduce the problem (e.g. with gdb or an extra sleep() somewhere) reliably?

adunstan:
To recreate the original problem, make a pretty large table. I used this table with 100m rows:

create table testme as select x,y from generate_series(1,10000) x, generate_series(10001,20000) y;
alter table testme add primary key(x,y);

Then set up this in file locker.sql:

begin;
select pg_sleep(10);
lock table testme;
rollback;

In session 1 do:

pg_repack -t testme -n -e mydb

In session 2, do 

psql -e -f locker.sql mydb

Session 1 will be blocked when it comes time to create the new index. Session 2 will be blocked waiting for pg_repack to finish.

The first commit on the indfix branch fixed this problem, but caused another, as pg_repack would be blocked when there was a pre-existing ACCESS EXLUSIVE lock on the table. The second commit avoids this problem while still fixing the first, by ensuring that it gets the index definitions when pg_repack already has an ACCESS EXCLUSIVE lock on the table.

schmiddy:
Really sorry I haven't gotten a chance to dig into this yet -- maybe this weekend.

schmiddy:
I was able to verify the deadlock behavior in pg_repack head using Andrew's testcase, and that this should fix the problem. Thanks for the fix!

Issue #16, opened 2013-11-05
Add (preliminary) support for PG 9.4


dvarrazzo:
From Thom Brown, after releasing 1.2.0-beta1:

    Just tried this out on 9.4 and I get the following error:

    psql (9.4devel)
    Type "help" for help.

    pgbench=# create extension pg_repack;
    ERROR:  column "reltoastidxid" does not exist at character 5461
    STATEMENT:  create extension pg_repack;
    ERROR:  column "reltoastidxid" does not exist

    This is due to the column being removed from pg_class as toast indexes
    will now be treated like regular indexes.  Could you look at adding
    support for this change?

dev1ant:
I have made an ugly fix to work around the absence of `reltoastidxid` field in `pg_class`. The patch can be taken [here](http://simply.name/tmp/0001-Small-fix-for-compability-with-9.4.patch). It compiles and extension installs but it seems, that there is one more problem with 9.4 compability. Calling `repack.disable_autovacuum` function with any oid leads to death of backend process with segfault:
```
< 2015-01-06 22:34:14.128 MSK >LOG:  server process (PID 20313) was terminated by signal 11: Segmentation fault
< 2015-01-06 22:34:14.128 MSK >DETAIL:  Failed process was running: select repack.disable_autovacuum(148806);
```

The backtrace of PID 20313 is:
```
(gdb) bt
#0  0x00007fd598f20e3c in vfprintf () from /lib64/libc.so.6
#1  0x00007fd598fd8aa0 in __vsnprintf_chk () from /lib64/libc.so.6
#2  0x0000000000794844 in vsnprintf (buf=0x19bbcc1 "ALTER TABLE ", len=1975, fmt=0x7fd373b37300 "ALTER TABLE %s SET (autovacuum_enabled = off)", args=0x7ffff42f0110) at /usr/include/bits/stdio2.h:78
#3  pvsnprintf (buf=0x19bbcc1 "ALTER TABLE ", len=1975, fmt=0x7fd373b37300 "ALTER TABLE %s SET (autovacuum_enabled = off)", args=0x7ffff42f0110) at psprintf.c:121
#4  0x00000000005bb3ff in appendStringInfoVA (str=0x7ffff42f00f0, fmt=<value optimized out>, args=<value optimized out>) at stringinfo.c:130
[...]

PR #33, opened 2015-01-07
Ugly fix to support PostgreSQL 9.4

dev1ant:
This fix does not have support for multiple TOAST indexes (which may happen with PostgreSQL 9.4 and higher) but at least right now pg_repack could be installed on PostgreSQL 9.4 and regression tests passes. See [this issue](https://github.com/reorg/pg_repack/issues/16) for more details.
If I understand everything right, for full support of several toast indexes function `repack_swap` should be strongly rewritten. Unfortunately, I'm not able to do it.


dev1ant:
Actually, I hurried up. This patch breaks compatibility with older versions.

PR #34, opened 2015-01-12
Ugly fix to support PostgreSQL 9.4

dev1ant:
First try was in [this PR](https://github.com/reorg/pg_repack/pull/33). Now it passes regression tests on both PostgreSQL 9.3 and PostgreSQL 9.4.

### DIFF
--- a/bin/pgut/pgut.c
+++ b/bin/pgut/pgut.c
@@ -1210,6 +1210,11 @@ exit_or_abort(int exitcode)
 
 /*
  * unlike the server code, this function automatically extend the buffer.
+ *
+ * TODO: as of http://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=3147acd6
+ * server's API of appendStringInfoVA changed. Should be done the same here
+ * or should this function be removed?
+ * 
  */
 bool
 appendStringInfoVA(StringInfo str, const char *fmt, va_list args)
--- a/lib/pg_repack.sql.in
+++ b/lib/pg_repack.sql.in
@@ -172,7 +172,7 @@ CREATE VIEW repack.tables AS
   SELECT R.oid::regclass AS relname,
          R.oid AS relid,
          R.reltoastrelid AS reltoastrelid,
-         CASE WHEN R.reltoastrelid = 0 THEN 0 ELSE (SELECT reltoastidxid FROM pg_class WHERE oid = R.reltoastrelid) END AS reltoastidxid,
+         CASE WHEN R.reltoastrelid = 0 THEN 0 ELSE (SELECT indexrelid FROM pg_index WHERE indrelid = R.reltoastrelid) END AS reltoastidxid,
          N.nspname AS schemaname,
          PK.indexrelid AS pkid,
          CK.indexrelid AS ckid,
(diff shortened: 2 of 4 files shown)


schmiddy:
Thanks, will try to poke into this as soon as we get 1.3.0 out the door.

NikolayS:
Any news about it? 9.4.1 is OU already

NikolayS:
*out

kmoppel:
+1 :)

schmiddy:
Definitely haven't forgotten about this, just haven't had time to make sure this self-described "ugly fix" is OK to go into master. And thanks @dev1ant for putting this PR together and everyone's interest in supporting 9.4.

I wanted to audit the toast index change in 9.4 a bit more closely to make sure there were no surprises in store. For instance, are there any regression tests which would be useful safeguards to add here? Can we check that we really only have one toast index to rebuild and swap instead of possibly multiple indexes which IIRC are theoretically supported now (or may be in a future PG release)?

dev1ant:
Well, this fix is ugly because it indeed creates potential problems in the future.To add support for 9.4 beautifully and cleanly much more work (as you mentioned above) should be done. But I have already [written](https://github.com/reorg/pg_repack/issues/16#issuecomment-68391560) that I'm not able to do it :( I've made a patch that works for me right now in a hope that it will help you to make a good fix.

IMHO the code of pg_repack should be divided into branches for different versions of PostgreSQL. Because in `lib/pg_repack.sql.in` for example you can't write different queries for different version of PostgreSQL. And lots of `#if PG_VERSION_NUM ...` would be removed and code would be easier to read. But that is really a big change in codebase and more work in maintenance of that code.

dvarrazzo:
Conditional support can be introduced no problem from Postgres 9.0 (e.g. using DO and plpgsql, which is installed by default in PG 9.0 servers)

I'm reviewing the 9.4 support and I think we can more easily support it by dropping support for PG 8.4, which is now dropped out of support. So I would release it as pg_repack 1.4. @schmiddy is that ok for you?

schmiddy:
Yeah, I think that's alright given that 8.4 is EOL'ed now. IIRC I pushed for us to drop 8.3 support some time ago.

dvarrazzo:
Sweet, I'm working on it right now as the pg_reorg 1.1.11 implementation is worrying.... :\

dvarrazzo:
Actually, their implementation is not that bad: AFAICS navigating table.reltoastrelid -> toast.oid -> index.indrelid works for PG < 9.4 too (at least I've checked it on PG 9.3), as well as the now unsupported table.reltoastrelid -> toast.reltoastidxid -> index.oid. With an important difference that can be spotted in [the server changeset](https://github.com/postgres/postgres/commit/2ef085d0e6960f5087c97266a7211d37ddaa9f68#diff-1fb3532a9acf402d7e5a523b569f71e4R1295): there could be multiple indexes but only one valid. So apparently pg_reorg 1.1.11 is not broken for PG < 9.4 but still could be wrong on PG 9.4.

If we can support PG 9.4 without introducing 9.0-only dependencies I'd go for a release of pg_repack 1.3.1 instead.

dvarrazzo:
Better implemented in <https://github.com/dvarrazzo/pg_repack>.

dev1ant:
Cool. Thanks!

dvarrazzo:
@dev1ant thank you: mine are mostly adjustments on top of your work ;)

PR #35, opened 2015-01-14
Add a expected out file to run the repack regression test successfully.

bwtakacy:
The 'repack' regression test fails on PostgreSQL 8.4 because the index name
assigned implicitely is different from 9.0 and after.

#### Probelm
```
$ cat regress/regression.diffs
*** /home/postgres/pg_repack/pg_repack/regress/expected/repack.out      2015-01-14 15:17:51.569064746 +0900
--- /home/postgres/pg_repack/pg_repack/regress/results/repack.out       2015-01-14 15:44:21.401063661 +0900
***************
*** 325,331 ****
  -- => OK
  \! pg_repack --dbname=contrib_regression --table=tbl_pk_uk --only-indexes
  INFO: repacking indexes of "tbl_pk_uk"
! INFO: repacking index "public"."tbl_pk_uk_col2_col1_key"
  INFO: repacking index "public"."tbl_pk_uk_pkey"
  -- => OK
  \! pg_repack --dbname=contrib_regression --table=tbl_nn_puk
--- 325,331 ----
  -- => OK
  \! pg_repack --dbname=contrib_regression --table=tbl_pk_uk --only-indexes
  INFO: repacking indexes of "tbl_pk_uk"
! INFO: repacking index "public"."tbl_pk_uk_col2_key"
  INFO: repacking index "public"."tbl_pk_uk_pkey"
  -- => OK
  \! pg_repack --dbname=contrib_regression --table=tbl_nn_puk

======================================================================
```

This diffs caused by the difference behaviors of PostgreSQL server to assign a index name.
In PostgreSQL 8.4,
```
postgres=# CREATE TABLE tbl_pk_uk (col1 int NOT NULL, col2 int NOT NULL, PRIMARY KEY(col1, col2), UNIQUE(col2, col1));
NOTICE:  CREATE TABLE / PRIMARY KEY will create implicit index "tbl_pk_uk_pkey" for table "tbl_pk_uk"
NOTICE:  CREATE TABLE / UNIQUE will create implicit index "tbl_pk_uk_col2_key" for table "tbl_pk_uk"
CREATE TABLE
```
but in PostgreSQL 9.0 and after,
```
postgres=# CREATE TABLE tbl_pk_uk (col1 int NOT NULL, col2 int NOT NULL, PRIMARY KEY(col1, col2), UNIQUE(col2, col1));
NOTICE:  CREATE TABLE / PRIMARY KEY will create implicit index "tbl_pk_uk_pkey" for table "tbl_pk_uk"
NOTICE:  CREATE TABLE / UNIQUE will create implicit index "tbl_pk_uk_col2_col1_key" for table "tbl_pk_uk"
CREATE TABLE
```

#### Fix
I think there are two way to fix this.
One is to modify Makefile which check PostgreSQL server version and convert the index name in regress/expected/repack.out.
Another is to add one more expected file, regress/expected/repack_1.out which correspond the output against PostgreSQL 8.4.

### DIFF
--- a/regress/expected/repack_1.out
+++ b/regress/expected/repack_1.out
@@ -1,0 +1,389 @@
+SET client_min_messages = warning;
+--
+-- create table.
+--
+CREATE TABLE tbl_cluster (
+	col1 int,
+	"time" timestamp,
+	","")" text,
+	PRIMARY KEY (","")", col1) WITH (fillfactor = 75)
+) WITH (fillfactor = 70);
+CREATE INDEX ","") cluster" ON tbl_cluster ("time", length(","")"), ","")" text_pattern_ops) WITH (fillfactor = 75);
+ALTER TABLE tbl_cluster CLUSTER ON ","") cluster";
+CREATE TABLE tbl_only_pkey (
+	col1 int PRIMARY KEY,
+	","")" text
+);
+CREATE TABLE tbl_only_ckey (
+	col1 int,
+	col2 timestamp,
+	","")" text
+) WITH (fillfactor = 70);
+CREATE INDEX cidx_only_ckey ON tbl_only_ckey (col2, ","")");
+ALTER TABLE tbl_only_ckey CLUSTER ON cidx_only_ckey;
+CREATE TABLE tbl_gistkey (
+	id integer PRIMARY KEY,
+	c circle
+);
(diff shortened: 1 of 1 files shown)


schmiddy:
Thanks for the fix, and sorry for the delay.

PR #36, opened 2015-01-16
Improve error message when repack against non-existing table.

bwtakacy:
Previously, pg_repack shows "ERROR: ERROR: relation foo does not exist" when specify non-existing table. 
Though the first ERROR is from pg_repack and the second ERROR is from PostgreSQL server, some users might think that pg_repack shows error level twice wrongly. So, this patch adds message explain it.


schmiddy:
Thanks for the fix, committed with a little wordsmithing. It is kind of a nuisance that we end up with two "ERROR:" strings in the output, would be nice to actually fix that, but this'll do for now.

PR #41, opened 2015-04-17
Patch for self cleaning process in pg_repack on error.

kotsachin:
This patch contains repack_cleanup_callback() which calls to repack_drop() for cleaning temporary objects. repack_cleanup_callback() will be pushed on stack using pgut_atexit_push() at beginning so that it will pop on abort or exit of program. 
This patch includes one global counter ("temp_obj_num") which counts number of temporary objects created by pg_repack. Correct order of deletion of temporary object as per count avoids unintentional error messages.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -243,7 +243,7 @@ static char *
 utoa(unsigned int value, char *buffer)
 {
 	sprintf(buffer, "%u", value);
-	return strdup(buffer);
+	return pgut_strdup(buffer);
 }
 
 static pgut_option options[] =
@@ -1391,6 +1391,7 @@ repack_one_table(repack_table *table, const char *orderby)
 	params[1] = utoa(temp_obj_num, buffer);
 	command("SELECT repack.repack_drop($1, $2)", 2, params);
 	command("COMMIT", 0, NULL);
+	temp_obj_num = 0; /* reset temporary object counter after cleanup */
 
 	/*
 	 * 7. Analyze.
@@ -1706,6 +1707,7 @@ repack_cleanup_callback(bool fatal, void *userdata)
 
 		reconnect(ERROR);
 		command("SELECT repack.repack_drop($1, $2)", 2, params);
+		temp_obj_num = 0; /* reset temporary object counter after cleanup */
 	}
 }
 
@@ -1734,6 +1736,7 @@ repack_cleanup(bool fatal, const repack_table *table)
 		params[0] = utoa(table->target_oid, buffer);
 		params[1] =  utoa(temp_obj_num, buffer);
 		command("SELECT repack.repack_drop($1, $2)", 2, params);
+		temp_obj_num = 0; /* reset temporary object counter after cleanup */
 	}
 }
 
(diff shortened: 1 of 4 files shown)


schmiddy:
Agreed that it'd be nice if we could cut out the few cases left where users are stuck having to run `DROP EXTENSION pg_repack CASCADE;`, so thanks for putting this together. If I remember the history of that code, we used to have an atexit handler performing this cleanup which we inherited from the pg_reorg code, but then moved to the current repack_cleanup() function for simplicity.

In hindsight, maybe it is easier to just use the atexit handler as this patch does. A few questions/comments about this PR though:

 * I see you're using pgut_atexit_push(). There's no need to call pgut_atexit_pop()?
 * Any way to consolidate repack_cleanup() and repack_cleanup_callback(), they seem duplicative now?
 * Any reason for the unconditional reconnect() in repack_cleanup_callback() whereas repack_cleanup() uses tests of PQstatus?
 * Better to use pgut_strdup() instead of strdup() (N.B. a tiny memory leak now)
 * This handler is supposed to work if the pg_repack client gets SIGINT'ed, right? Note, I seem to fall into an infinite loop of repack_cleanup_callback() when I try to Control-C the pg_repack client while it is rebuilding a table.

kotsachin:
Please find my comments as below:

>In hindsight, maybe it is easier to just use the atexit handler as this patch does. A few questions/comments about this PR though:

>I see you're using pgut_atexit_push(). There's no need to call pgut_atexit_pop()?

Yes. In this case there is no need to call pgut_atexit_pop().
As we pushed repack_cleanup_callback() on stack initially it will be automatically pop in call_atexit_callbacks() at exit or abort. 


>Any way to consolidate repack_cleanup() and repack_cleanup_callback(), they seem duplicative now?

Yes. They are seems duplicate but they are handling exactly opposite scenarios.
If there is "fatal" repack_cleanup() simply display error and exit.
On "fatal" repack_cleanup_callback() clean the temporary objects created. 
If we use only one function for both purposes there is chance that it might go in infinite loop.

>Any reason for the unconditional reconnect() in repack_cleanup_callback() whereas repack_cleanup() uses tests of PQstatus?

I tried to use PQstatus for reconnect. It throws "object not found" error at PostgreSQL side. 
That is why I decided to use unconditional reconnect. I think it is unable to continue with dirty connection.
Error looks like following:
```
DETAIL: query was: SELECT repack.repack_drop($1, $2)
DETAIL: query was: SELECT repack.repack_drop($1, $2)
Segmentation fault (core dumped)
```

>Better to use pgut_strdup() instead of strdup() (N.B. a tiny memory leak now)

Yes. agree with you. I have updated this in newer commit.


>This handler is supposed to work if the pg_repack client gets SIGINT'ed, right? Note, I seem to fall into an infinite loop of >repack_cleanup_callback() when I try to Control-C the pg_repack client while it is rebuilding a table.

Yes. I found the same.
It is because my previous logic was little wrong. In that code call_atexit_callbacks() get executed with true value in all cases. This should not happen. I changed code so that now it will execute only on cleanup case.

schmiddy:
Thanks for the fixes, this seems to work reasonably well now in my testing. One bit that's not quite working is SIGINT handling during the rebuild-indexes phase with jobs > 1, i.e. with multiple worker conns. The problem predates patch -- I think it could  be mostly fixed by just taking out the check of `errno != EINTR` around line 916 of bin/pg_repack.c:

    if (ret < 0 && errno != EINTR)
        elog(ERROR, "poll() failed: %d, %d", ret, errno);

though in some testing I did, the on_interrupt() handler in bin/pgut/pgut.c didn't cancel the in-progress index builds. I'm a bit suspicious of the cancel coding in `on_interrupt()`, seems like it should be using `PQgetCancel` instead of monkeying around with `c->cancel`, but I haven't had a chance to try a fix yet.

PR #42, opened 2015-04-21
Adding --delete-key option to pg_repack.

mr-mustash:
This uses a WHERE clause during the copy stage, effectively doing a bulk delete. To implement this before I've directly modified the repack.tables view to have the "WHERE" clause that will cause the delete. This seemed hacky to me, so I've implemented it as an option in pg_repack. Below is an example of my implementation.

```
pking@[local]:5432 [pking]=# \d inbox_messages
                                      Table "public.inbox_messages"
   Column    |            Type             |                          Modifiers
-------------+-----------------------------+-------------------------------------------------------------
 id          | integer                     | not null default nextval('inbox_messages_id_seq'::regclass)
 listener_id | integer                     |
 expire_time | timestamp without time zone | 
 payload     | text                        | 
Indexes:
    "inbox_messages_pkey" PRIMARY KEY, btree (id)

pking@[local]:5432 [pking]=# select count(1) from inbox_messages ;
 count 
-------
   741
(1 row)

13:29:25 - [pking@pking-m02] {master} pg_repack $ ./bin/pg_repack --dbname="pking" --table=inbox_messages --order-by="listener_id" --delete-key="expire_time > now()" --port=5432 --echo

pking@[local]:5432 [pking]=# select count(1) from inbox_messages ;
 count 
-------
   510
(1 row)
```


dvarrazzo:
I wonder if it would be conceivable to have a update mechanism more generic than just delete. I'd like to have a discussion about the topic.

mr-mustash:
I agree that having a generic update mechanism would be better than just having the delete. The reason I've only implemented the delete is that it fits my usecase. It also has no performance impact since it happens during the copy phase.

dvarrazzo:
So we can leave this issue here as documentation of an use case but this issue is NOT to be merged.

andreasscherbaum:
@dvarrazzo already said that this will not be merged. Closing this (the history stays intact).

PR #46, opened 2015-05-13
Describe pg_repack behavior

### LINKED ISSUE #45: Document pg_repack behavior
Hi,

while looking for a documentation of what pg_repack does internally I stumbled over http://www.depesz.com/2011/07/06/bloat-happens/

In there the @depesz writes:

> Full procedure looks like this:
> 
> * create log table for changes
> * create triggers on old table to log changes
> * create new table with copy of all data in old table
> * create all indexes on new table
> * apply all changes from log table to new table
> * switch (in system catalogs) information about table files
> * drop old table

Is this still true for pg_repack? If so, it would be great if this could be mentioned in the docs.

terrorobe:
Fixes #45

### DIFF
--- a/doc/pg_repack.rst
+++ b/doc/pg_repack.rst
@@ -429,13 +429,35 @@ earlier versions which could result in data corruption.
 Details
 -------
 
-To perform full table repacks, pg_repack creates a work table in the "repack"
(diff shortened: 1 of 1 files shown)


schmiddy:
Thanks! Merged in with some tidying.

PR #47, opened 2015-05-27
Get rid of hacky usage of pgut_strdup() in utoa.

amitlan:
From inspecting the call sites of utoa, it appears that some of them
(especially the recent cleanup patch which added the strdup there) wanted
to prevent overwriting a local variable by repeated call (to utoa) using
the same variable as argument. This commit instead makes such call sites
strdup the variable itself before passing it to utoa. That seems cleaner
considering that it does not seem utoa's contract to do so (strdup its
parameter that is).

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -243,13 +243,8 @@ static char *
 utoa(unsigned int value, char *buffer)
 {
 	sprintf(buffer, "%u", value);
-	/* XXX: originally, we would just return buffer here without
-	 * the pgut_strdup(). But repack_cleanup_callback() seems to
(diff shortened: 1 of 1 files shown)


amitlan:
By the way, I apologize for not having commented on the commit introducing strdup in utoa(). It'd have saved shmiddly some trouble of having to guess what it did there.

schmiddy:
Ah thank you, I realize now what was going on with that extra pgut_strdup() inside utoa(). I like it a little better with char arrays that we don't need to worry about freeing, so fixed it up that way.

amitlan:
Thanks!

PR #52, opened 2015-10-11
repack_apply(): Use default if count is negative

decibel:
Per documentation, a negative value for count should result in using DEFAULT_PEEK_COUNT.


ooyamams:
Hi, I found this small bug and quickly fixed.
Thanks.

Issue #55, opened 2015-11-02
Deadlock detected after pg_repack receives SIGINT


jirihlinka:
I'm running a pg_repack from a bash script with timeout of 10 minutes like so (simplified version):
timeout -s SIGINT 10m pg_repack --table=frequently_updated_table

After the 10 min timeout, the OS sends SIGINT to pg_repack process so the pg_repack calls:
SELECT repack.repack_drop($1, $2)
and it causes a deadlock with other process which is INSERTing into frequently_updated_table that has a pg_repack's trigger to log every change to a repack.log_[OID of frequently_updated_table].

Is it a bug of pg_repack or am I doing something wrong? Should I add some detailed info?

Thank You for any help,
Jiri

Update: pg_repack version 1.3.2

amitlan:
It's highly likely that it's a bug of repack.repack_drop().

Could you check if you the issue is resolved with 5317f527f4a502347782030e30ca05db83c57584?

jirihlinka:
Thanks Amit for reply. May I kindly ask You for a help (or link to how-to
:-)) with how to try the 5317f52? I'm using RPM from yum.postgresql.org
currently (
http://yum.postgresql.org/9.2/redhat/rhel-6-x86_64/pg_repack92-1.3.2-1.rhel6.x86_64.rpm
).

On Fri, Nov 6, 2015 at 9:18 AM, Amit Langote <notifications@github.com>
wrote:

> It's highly likely that it's a bug of repack.repack_drop().
>
> Could you check if you the issue is resolved with 5317f52
> <https://github.com/reorg/pg_repack/commit/5317f527f4a502347782030e30ca05db83c57584>
> ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/reorg/pg_repack/issues/55#issuecomment-154341817>.
>



-- 
Bc. Jiří Hlinka
Tel.: 725 315 263

amitlan:
@jirihlinka you can clone the repository [here](https://github.com/amitlan/pg_repack.git) and build pg_repack from the branch issue-55-fix. Have you tried building pg_repack from source before? Something like below?

```
cd /tmp
git clone https://github.com/amitlan/pg_repack.git
cd pg_repack
git checkout issue-55-fix
make install
```

jirihlinka:
@amitlan : Great, thanks for help.

[...]

PR #56, opened 2015-11-06
Make repack_drop() processing robust against deadlocks.

amitlan:
Concurrent activity on the target table can cause deadlocks when
repack_drop() is doing its job, ie, dropping the temporary objects
created. It is highly likely to occur when pg_repack is interrupted
midway through its processing.

This fixes deadlock reported in issue #55 .

### DIFF
--- a/lib/repack.c
+++ b/lib/repack.c
@@ -968,8 +968,8 @@ repack_drop(PG_FUNCTION_ARGS)
 	 */
 	execute_with_format(
 		SPI_OK_UTILITY,
-		"LOCK TABLE %s IN ACCESS EXCLUSIVE MODE",
-		relname);
+		"LOCK TABLE %s.%s IN ACCESS EXCLUSIVE MODE",
+		nspname, relname);
 
 	/* drop log table: must be done before dropping the pk type,
 	 * since the log table is dependent on the pk type. (That's
--- a/lib/repack.c
+++ b/lib/repack.c
@@ -950,8 +950,19 @@ repack_drop(PG_FUNCTION_ARGS)
 	repack_init();
 
 	/*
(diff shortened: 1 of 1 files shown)


schmiddy:
Well, no objections to this fix, but can anyone verify whether this actually solves Issue #55 (and hopefully suggest how I could reproduce that problem myself)?

amitlan:
To reproduce the deadlock with HEAD - create a big enough pgbench database (say, with scale 50); start a pgbench test; in another terminal do -

timeout -s SIGKILL -t 5s pg_repack -t pgbench_accounts -d pgbench

I sent the PR after making sure the deadlock I saw before disappeared once patched. But see if you can reproduce it before looking at the patch if it's really a "complete" fix.

amitlan:
Sorry that was SIGINT, not SIGKILL.

amitlan:
@schmiddy I have expanded the comment in repack_drop() to clarify why a deadlock could occur. See if that makes sense.

schmiddy:
OK, I think I've reproduced the deadlock, looks something like this for the pgbench example:
```

amitlan:
Oops, totally missed that. Will look into it.

amitlan:
Apparently, I failed to include schema name in the LOCK TABLE command. I pushed a fix for the same. That fixes the regression test failure. I should have checked them already.

schmiddy:
OK, this seems good now. (We should probably still fix that infinite loop upon error in repack_drop, but that's a separate issue.)

amitlan:
Thanks Josh! I agree it would be nice to fix that infinite loop thingy. Will see if I can do something about that if get time.

PR #57, opened 2015-11-16
Escape table name in repack_table_indexes

mahlberg:
Fixes issues when trying to do an index only replack where the table name needs to be escaped, e.g. if the table is named user or order.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -1888,7 +1888,7 @@ repack_table_indexes(PGresult *index_details)
 
 	/* take an exclusive lock on table before calling repack_index_swap() */
 	resetStringInfo(&sql);
-	appendStringInfo(&sql, "LOCK TABLE %s IN ACCESS EXCLUSIVE MODE",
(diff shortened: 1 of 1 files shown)


schmiddy:
So the big problem here is that we're mostly trying to mimic the interface of other PostgreSQL-provided utilities particularly `clusterdb` or `vacuumdb`. Here's a fun example to try:

    clusterdb -t "mytable; DROP TABLE baloney;"

Obviously, don't try that if you have a table name `baloney` in your database -- I think you can tell what will happen. AFAICT the expectation from PostgreSQL is that you as the user are in charge of quoting/escaping your table names so that `clusterdb` or `vacuumdb` doesn't have to. Yes, this is a stupid expectation. But that's the example we are following :-(

(That said, it's been a while since I've tested needing-to-be-quoted table names with `pg_repack` -- if we're not consistent with `clusterdb` / `vacuumdb`'s behavior of allowing the user to pass in an already-quoted table name, we should fix that.)

mahlberg:
You're right, it seems to be the same behaviour with ``vacuumdb`` and I agree that it makes sense for these tools to be consistent. Thanks for the quick and clear answer.

PR #61, opened 2015-12-07
Add doc in Japanese

bwtakacy:
This is mostly for users in Japan who currently use pg_reorg but want to move to pg_repack.
The original text is kept as comment.

Translated by Takashi Ohnishi (bwtakcy).


schmiddy:
Thanks!

PR #64, opened 2016-01-18
Need to #include pg_am.h directly now to use BTREE_AM_OID.

amitlan:
Due to upstream changes (see commit 65c5fcd3 in HEAD) to the index AM API definitions.


amitlan:
I'm really sorry Daniel, I accidentally caused your comment on a commit to be deleted because I force pushed a new commit. I agree that a version include guard is better so added in the latest commit.

PR #65, opened 2016-01-18
Fix regression test output

amitlan:
Upstream changes (see commit c7e27bec in HEAD) now cause reloption values to be emitted with surrounding quotes. That breaks, for example, the outputs of \d commands in one of the tests.


schmiddy:
Thanks for the fix!

PR #66, opened 2016-01-18
Allow wildcard in options

isaiasanchez:
For example we have some partitioned tables using same prefix, would be nice if we can repack these set of tables using something like:

pg_repack -t myprefix* -d mydb

Same way as pg_dump.


schmiddy:
Hi @isaiasanchez, it is possible we could consider this as a feature request, but this Pull Request seems to be malformed -- it is asking to merge 71 commits from our `maint_1.1` branch into our `master` branch.

FWIW you may want to propose the feature by opening an Issue first before coding it up -- we generally try to keep the behavior of the `pg_repack` client binary roughly similar to how `clusterdb` and `vacuumdb` work, and I don't think those support wildcard specification of the `--table` option.

PR #72, opened 2016-02-12
Do not repack unlogged tables (close #71)

### LINKED ISSUE #71: pg_repack corrupts an unlogged table
I have no idea how it happend, but this is what I seen:

   1. try to pg_repack an unlogged table
   2. postgresql killed by linux oom-killer
   3. pg_repack try to reconnect, but failed because postmaster is recovering
   4. after recovery, pg_repack continue (?) and run  repack.repack_drop
   5. my application is restarted.
   6. postmaster complain the index have zero page, suggest me to reindex
   7. reindex the table give duplicated rows and fails.

I guess unlogged tables should not be repacked in the first place, right?

j16sdiz:
Fix (#71).


schmiddy:
This will do for the moment. It'd be nicer to have an error message telling the user what's wrong, the way we do for:
```
"relation \"%s\" must have a primary key or not-null unique keys"
```

PR #87, opened 2016-07-08
Don't wait for concurrent vacuums

akorotkov:
Our customers were faced following issue: pg_repack waits for concurrent vacuum to be complete before swapping tables. Vacuum could be slow on very large tables. Therefore it appears to be real problem.
Since VACUUMs are always outside transaction block, there is no risk that their sessions could see old data. This is why we propose to exclude vacuums (both autovacuums and manual vacuums) from list of vxids to wait.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -66,6 +66,7 @@ const char *PROGRAM_VERSION = "unknown";
  *     https://github.com/reorg/pg_repack/issues/1
  *  d. open transactions/locks existing on other databases than the actual
  *     processing relation (except for locks on shared objects)
+ *  e. VACUUMs which are always executed outside transaction blocks.
  *
  * Note, there is some redundancy in how the filtering is done (e.g. excluding
  * based on pg_backend_pid() and application_name), but that shouldn't hurt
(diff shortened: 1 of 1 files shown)


jwilm:
Thanks for putting this together, @akorotkov! We've had the same problem; glad to see there's a fix on the way. 😄

akorotkov:
Hi! Glad to see this pull request is well accepted. But what prevents us from committing this?
We in PostgresPro are also interested in pg_repack development. If you grant me committing rights to this repository, I can participate in maintenance as well.

bin/pg_repack.c
```diff
@@ -84,6 +85,8 @@ const char *PROGRAM_VERSION = "unknown";
 	"  AND l.pid NOT IN (pg_backend_pid(), $1) " \
 	"  AND (l.virtualxid, l.virtualtransaction) <> ('1/1', '-1/0') " \
 	"  AND (a.application_name IS NULL OR a.application_name <> $2)" \
+	"  AND a.query !~* E'^\\\\s*vacuum\\\\s+' " \
```
schmiddy: Note it's possible there may be no trailing space after 'vacuum' in the query text, e.g. `VACUUM(freeze)mytable;` or just `VACUUM;`.

schmiddy:
@akorotkov thanks for this fix, I think this is reasonable.  As you can probably tell we certainly could use some help with maintenance.

akorotkov:
Great! Thanks!

PR #97, opened 2016-10-24
new option 'include-extensions' (don't touch tables that belong to extensions)

funbringer:
Hi guys,

I'm one of the developers of [pg_pathman](https://github.com/postgrespro/pg_pathman), a new partitioning tool for PostgreSQL.

Recently, one of our clients had trouble running pg_repack on a DB containing some tables partitioned by pg_pathman. In other words, pg_repack would hang _every time_ when trying to lock `pathman_config`, a table that we use to store partitioning configuration. As it turns out, when pg_repack opens a secondary connection and performs any query, our `post_parse_analyze_hook` comes into play and tries to read `pathman_config` (for planning purposes), which is impossible, hence the deadlock :(

This gave me an idea that it would be nice if we could make pg_repack forget about `pathman_config` and other tables that belong to their extensions. There might always be some code that heavily depends on such tables, so I've created a new option called `include-extensions` (which is set to false by default) to include them when it's really needed.

Hope this works out.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -625,6 +625,12 @@ repack_one_database(const char *orderby, char *errbuf, size_t errsize)
 	params[iparam++] = tablespace;
 	if (num_tables)
 	{
+		/*
+		 * Tables have been explicitly specified,
+		 * no need to hide tables of extensions
+		 */
+		include_extensions = true;
+
 		appendStringInfoString(&sql, "(");
 		for (cell = table_list.head; cell; cell = cell->next)
 		{
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -251,7 +251,7 @@ static bool				dryrun = false;
 static unsigned int		temp_obj_num = 0; /* temporary objects counter */
 static bool				no_kill_backend = false; /* abandon when timed-out */
 static bool				no_superuser_check = false;
-static bool				include_extensions = false; /* repack tables of extensions */
+static SimpleStringList	exclude_extension_list = {NULL, NULL}; /* don't repack tables of these extensions */
 
 /* buffer should have at least 11 bytes */
 static char *
@@ -279,7 +279,7 @@ static pgut_option options[] =
 	{ 'i', 'j', "jobs", &jobs },
 	{ 'b', 'D', "no-kill-backend", &no_kill_backend },
 	{ 'b', 'k', "no-superuser-check", &no_superuser_check },
-	{ 'b', 'C', "include-extensions", &include_extensions },
+	{ 'l', 'C', "exclude-extension", &exclude_extension_list },
 	{ 0 },
 };
 
@@ -607,12 +607,14 @@ repack_one_database(const char *orderby, char *errbuf, size_t errsize)
 	size_t					num_tables;
 	size_t					num_schemas;
 	size_t					num_params;
+	size_t					num_excluded_extensions;
 
 	num_tables = simple_string_list_size(table_list);
 	num_schemas = simple_string_list_size(schema_list);
+	num_excluded_extensions = simple_string_list_size(exclude_extension_list);
 
 	/* 1st param is the user-specified tablespace */
-	num_params = num_tables + num_schemas + 1;
+	num_params = num_excluded_extensions + num_tables + num_schemas + 1;
 	params = pgut_malloc(num_params * sizeof(char *));
 
 	initStringInfo(&sql);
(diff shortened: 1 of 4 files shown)


funbringer:
Anyone?

schmiddy:
Thanks @funbringer .

The original intent of the `pg_repack` client was to mimic the command-line tools from PG core like `clusterdb`, so I don't want to encourage too much drift, particularly changing the default behavior for deciding which tables to repack. That said, I think this is basically reasonable, and fits in with making the whole-database repack behavior Do the Right Thing.

funbringer:
@schmiddy, does this mean that you will accept my patch? Your answer looks rather vague to me. If so, do I have to make any adjustments regarding the default behavior?

schmiddy:
I will merge this if there are no objections.

funbringer:
> I will merge this if there are no objections.

Ok, thanks!

MichaelDBA:
Any regression testing done befor or after the merge?

schmiddy:
@MichaelDBA right, I like to test such changes myself before clicking merge (only reason I haven't merged yet). Our `installtest` tests today are far from adequate, I would love to have more comprehensive tests in general.

funbringer:
I hope you'll merge this some day :(

MasahikoSawada:
I'm not a member of pg_repack but here are some comments;

* Since the default value of `include_extensions` is false, the default behavior of pg_repack is going to be changed. To not confuse the user, is it better to set true by default?
* I think it's better to specify the extension name we want to skip rather than having bool.
* Documentation and regression test need to be updated.

maksm90:
Hi, @MasahikoSawada !

I want to comment on your first two points about details of specifying `include_extensions` option. We in `pg_pathman` team suppose that operations carrying out by `pg_repack` (lightweight re-clustering of clustered indexes and lightweight re-vacuum of tables) are redundant for the tables belonging to extensions because these tables generally store private information that IMHO is rarely updated. Moreover, as @funbringer has already mentioned, some extensions may suffer from such interference: in our case `pg_repack` simply hangs, and there's no way to prevent this situation. It goes without saying that `pg_repack` can't possibly handle each of these special cases.

In general, I agree that our design to exclude extension's tables may not be as flexible as yours. But we strongly believe that extensions' tables don't require `pg_repack`'s operations.

Could you provide an example of extension that would require such special treatment? I hardly can imagine that.

funbringer:
to @MasahikoSawada:

> Since the default value of include_extensions is false, the default behavior of pg_repack is going to be changed.

Is `pg_repack` supposed to repack tables that belong to extensions? It would still repack user-defined tables (i guess it's the main purpose), wouldn't it? If so, what's going to change?

> I think it's better to specify the extension name we want to skip rather than having bool.

I'm sure that current behavior regarding extensions is **incorrect**. That's why I created this pull request in the first place. I just wanted to provide a rock-steady way to prevent this kind of troubles from happening. Besides, it would still be possible to repack tables by explicitly specifying their names:

```shell
pg_repack -t table db
```

> regression test need to be updated.

It's hard to test this particular case, since that would require creating a toy extension containing a table. Is this really necessary? Otherwise, there are no observable deviations in `pg_repack`'s behavior.

Nonetheless, thanks for your review.

MasahikoSawada:
Basically I agree to add this feature but what I'm concerned is  to change default behavior of pg_repack.

> Could you provide an example of extension that would require such special treatment? I hardly can imagine that.

I guess that pg_dbms_stats, pg_paxos and so on could need such a special treatment. But since there are a lot of extensions in the world, there might be such a extension. Or such extension might come. It's the fact that pg_repack has been supposing to repack all tables in a database. I don't think it makes sense that all tables is not repacked even though the user specified all databases (-a option).

> It's hard to test this particular case, since that would require creating a toy extension containing a table. Is this really necessary? Otherwise, there are no observable deviations in pg_repack's behavior.

The regression test for every feature is necessary. But sorry, I came up with no good idea for it for now.

MasahikoSawada:
After thought, how about the following options?
1. Changing option name to `--exclude-extensions`. The default value is false. The tables belonging to extension are skipped.
2. Changing option name to `--exclude-schema`. The default value is NULL. The tables in specified schema are skipped. The tables belonging to extension mostly created on particular schema that extension created.

funbringer:
hi @MasahikoSawada,

> Changing option name to --exclude-extensions. The default value is false. The tables belonging to extension are skipped.

I think you might be right. I've been thinking for a while, and it seems that this might be a safer option. I'll make some adjustments.

funbringer:
to @MasahikoSawada:

I've changed both meaning and name of this option, now it's `--exclude-extension`:

```bash
./bin/pg_repack -cpublic --exclude-extension=pg_pathman debug
INFO: repacking table "test_hash"
INFO: repacking table "test_range"
INFO: repacking table "test_range_1"
INFO: repacking table "test_range_2"
```

Now it's much more flexible.

MasahikoSawada:
Great. The code looks good to me. I'll merge it after double checked.

dvarrazzo:
Please add a _what's new_ entry too to `pg_repack.rst` when merging, than you ;)

funbringer:
Thanks! I'll resolve conflicts with my other pull request ASAP.

Issue #106, opened 2017-01-19
Bug with INSERT CONFLICT


MasahikoSawada:
Hi,

I found that pg_repack can not be used together with INSERT CONFLICT. If INSERT CONFLICT DO NOTHING or DO UPDATE is executed during repacking the table, pg_repack fails with a high probability.

For example, a table with primary key on id column has a data `id = 1`, and a transaction executes `INSERT INTO test_table VALUES(1) CONFLICT ON (id) DO NOTHING` during repacking. The insertion will conflict on unique constraint and then do nothing. But insert operation is written to the operation log table. And then pg_repack fails in the apply phase. If `CONFLICT (id) DO UPDATE id = 2`, two records will be inserted to the operation log table, and then fails.

The solution I came up with is changing the trigger type from BEFORE to AFTER. We define AFTER trigger that is the first of the AFTER trigger to fire on the table. I believe that this change does not break existing functionality but if there is the specific reason why we have to define BEFORE trigger please give me feedback.

MasahikoSawada:
PR #107 has been merged.

PR #107, opened 2017-01-19
Change trigger type to AFTER trigger.

MasahikoSawada:
During repacking table, if a transaction executes INSERT CONFLICT
ON UPDATE/DO NOTHING, because we define BEFORE trigger on target
table, the contents of operation log table becomes inconsistent
easliy. As a result, pg_reapck fails with a high probability.

To resolve this issue, this changes the trigger type from BEFORE
to AFTER. We define AFTER trigger that is the first of the AFTER
trigger to fire on the table.

This fixes issue #106.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -186,8 +186,8 @@ typedef struct repack_table
 	Oid				ckid;			/* target: CK OID */
 	const char	   *create_pktype;	/* CREATE TYPE pk */
 	const char	   *create_log;		/* CREATE TABLE log */
-	const char	   *create_trigger;	/* CREATE TRIGGER a_repack_trigger */
-	const char	   *enable_trigger;	/* ALTER TABLE ENABLE ALWAYS TRIGGER a_repack_trigger */
+	const char	   *create_trigger;	/* CREATE TRIGGER repack_trigger */
+	const char	   *enable_trigger;	/* ALTER TABLE ENABLE ALWAYS TRIGGER repack_trigger */
 	const char	   *create_table;	/* CREATE TABLE table AS SELECT */
 	const char	   *drop_columns;	/* ALTER TABLE DROP COLUMNs */
 	const char	   *delete_log;		/* DELETE FROM log */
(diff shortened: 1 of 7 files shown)


schmiddy:
Good find, thanks for the fix. It's a significant change, but I'm not aware of any reason why pg_repack's trigger was declared as a BEFORE trigger instead of AFTER. Does anyone else see a problem with this change?

dvarrazzo:
It would also solve the problem of the trigger having to be the last one to trigger. I don't think I've ever found the reason for that choice, but there must be one for what looks a weirdness of the original design.

MasahikoSawada:
Thank you for comment!

That code has been existing since initial commit of pg_reorg. One possible reason is that BEFORE trigger is cheaper than AFTER trigger as [document](https://www.postgresql.org/docs/9.6/static/trigger-definition.html) says.

```
If you have no specific reason to make a trigger BEFORE or AFTER, the BEFORE case is more efficient, since the information about the operation doesn't have to be saved until end of statement.
```

> It would also solve the problem of the trigger having to be the last one to trigger.

I think that the trigger pg_repack defined have to be the first one AFTER trigger to fire on target table instead in order to avoid changing triggered tuple by another AFTER trigger.

dvarrazzo:
> I think that the trigger pg_repack defined have to be the first one AFTER trigger to fire on target table instead in order to avoid changing triggered tuple by another AFTER trigger.

If we move to AFTER triggers we should test it, but I don't think it would be required for the trigger to be the first. AFTER triggers cannot change the tuple written, they can only fire additional statements, which in turn may create new records, and maybe even in the same table, but these records would generate their own triggers.

MasahikoSawada:
> AFTER triggers cannot change the tuple written, they can only fire additional statements, which in turn may create new records, and maybe even in the same table, but these records would generate their own triggers.

Ah, you're right. The triggered tuple can not be changed. So can we remove test case and code for checking competing trigger?

dvarrazzo:
If we go for using AFTER TRIGGER (which I'm not sure 100% is a good idea... but can't find a counter-example and wouldn't know who to ask for one) yes, I'd be happy to see the brittle check for being last (or first) trigger gone altogether, and call it just `repack_trigger`.

MasahikoSawada:
Thank you for your advice. I've changed this PR. Please review it.

schmiddy:
@MasahikoSawada LGTM, I think this should be merged. It's a significant change and there is always some risk with big changes, but I think it's well worth it given that right now `INSERT ... ON CONFLICT` queries may blow up or cause silent corruption.

MasahikoSawada:
Merged, thanks. I'm going to continue testing it.

PR #108, opened 2017-01-24
Introduce --dont-kill-backend option.

MasahikoSawada:
pg_repack needs to take an exclusive lock at the end of the
reorganization. If the lock cannot be taken after duration
--wait-timeout option specified and this option is true,
pg_repack gives up to repack a target table instead of
cancelling conflicting backend. False by default.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -154,8 +154,8 @@ const char *PROGRAM_VERSION = "unknown";
 
 #define COUNT_COMPETING_LOCKS \
 	"SELECT pid FROM pg_locks WHERE locktype = 'relation'" \
-	"AND granted = false AND relation = %u" \
-	"AND mode = 'AccessExclusiveLock' AND pid <> pg_backend_pid()"
+	" AND granted = false AND relation = %u" \
+	" AND mode = 'AccessExclusiveLock' AND pid <> pg_backend_pid()"
 
 /* Will be used as a unique prefix for advisory locks. */
 #define REPACK_LOCK_PREFIX_STR "16185446"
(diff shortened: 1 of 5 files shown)


MasahikoSawada:
In current design, `lock_exclusive` and `kill_ddl` check the number of competing backend and then tries to cancel it if `dont-kill-backend` is not set. I think that there is still room to improve this feature and option name. Please test and give me feedback.

MichaelDBA:
What do you suggest for improvement? None seems apparent to me.

MasahikoSawada:
There are some points where pg_reapck tries to cancel other backend; lock_exclusive and kill_ddl. In current design pg_repack could give up repacking the table at everywhere. I'm not sure it's best behavior.

Also `kill_ddl` checks if there is conflicting backend and then does `pg_cancel_backend` and `pg_terminate_backend` immediately if there is. If `dont-kill-backend` is set, `kill_ddl` gives up doing that without waiting for finishing DDL. It should be separated issue but I think that we can make `kill_ddl` have timeout same as what `lock_exclusive` doing.

bin/pg_repack.c
```diff
@@ -152,6 +152,11 @@ const char *PROGRAM_VERSION = "unknown";
 	" AND granted = false AND relation = %u"\
 	" AND mode = 'AccessExclusiveLock' AND pid <> pg_backend_pid()"
 
+#define COUNT_COMPETING_LOCKS \
+	"SELECT pid FROM pg_locks WHERE locktype = 'relation'" \
+	"AND granted = false AND relation = %u" \
```
schmiddy: Looks like some whitespace is missing at the end of lines

MasahikoSawada:
Thank you for looking at this.
Fixed.

MasahikoSawada:
After thought, I'm thinking option name `no-kill-backend` is better than `dont-kill-backend` for consistency with other similar options such as `no-order` and `no-analyze`. Thought?

schmiddy:
👍 for calling it `--no-kill-backend`

MasahikoSawada:
I've changed it to `no-kill-backend`, and updated error message regarding timed out. Also updated Japanese documentation for this option.

I'll merge it barring any objection.

MasahikoSawada:
Merged. Thank you for reviewing!

PR #111, opened 2017-02-03
fix compilation with msvc

drakkan:
__attribute__ macro is not defined in msvc and it is not essential to
the implementation. All it does  is tell the compiler that this function
is similar to printf, and  expects a printf-like format string
So for MSVC we define __attribute__ as a macro that do nothing


dvarrazzo:
Cherry-picked, thank you.

PR #113, opened 2017-02-14
Allow RDS customers to use the pg_repack client.

bossartn:
The current client checks for superuser before attempting to
execute pg_repack commands.  In Amazon RDS, customers are given
access to a psuedo-superuser role called rds_superuser, but they
are not given access to superuser.  However, rds_superusers will
otherwise have the ability to execute pg_repack commands in RDS.

This change alters the client code to ensure that roles with
membership to the rds_superuser role pass the superuser checks.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -203,6 +203,7 @@ typedef struct repack_table
 
 
 static bool is_superuser(void);
+static bool is_rds_superuser(void);
 static void check_tablespace(void);
 static bool preliminary_checks(char *errbuf, size_t errsize);
 static void repack_all_databases(const char *order_by);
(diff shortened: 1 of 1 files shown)


dvarrazzo:
ISTM that when Amazon will introduce the Amazon ABC infrastructure you will have an abc_superuser, and then on the XYZ infrastructure you will have an xyz_superuser... Wouldn't it be better if we just have a flag `--no-superuser-check` that would make the `is_superuser()` function return true, and if something break it's the user's responsibility?

bossartn:
Yes, I agree that a `--no-superuser-check` flag could be a better long-term approach for the project.  I'll go ahead and code up this change and post it.

bossartn:
I've opened pull request #114 for the alternative approach.

brauliobo:
@bossartn so, will it work on RDS now?

mviau123:
Is this work prior to RDS officially supporting pg_repack extension?  Extensions needs to approved by RDS before can be added.  They currently don't.  See:
https://forums.aws.amazon.com/thread.jspa?threadID=242418

Is there news that they are starting to work on it?  It be great.  Any tips is appreciated.

PR #114, opened 2017-02-15
Introduce --no-superuser-check option.

bossartn:
The current client checks for superuser before attempting to
execute pg_repack commands. In Amazon RDS, customers are given
access to a psuedo-superuser role called rds_superuser, but they
are not given access to superuser. However, rds_superusers will
otherwise have the ability to execute pg_repack commands in RDS.

This change introduces the --no-superuser-check option in the
client code so that users can disable the client-side superuser
checks.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -277,7 +277,7 @@ static pgut_option options[] =
 	{ 'B', 'Z', "no-analyze", &analyze },
 	{ 'i', 'j', "jobs", &jobs },
 	{ 'b', 'D', "no-kill-backend", &no_kill_backend },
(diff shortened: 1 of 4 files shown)


bossartn:
I would appreciate any feedback that you have on this PR.  Thanks!

MasahikoSawada:
Thank you for the PR!
I think that we should have a regression test for this option.

bossartn:
Sounds good, I will write a regression test for this.

bossartn:
I've added a regression test to this PR.

bossartn:
I would appreciate any additional feedback you have on this PR.  Thanks!

PR #115, opened 2017-02-15
Fix 'tablespace' regression test for latest postgres minor releases

cpaelzer:
The recent stable update changed output for calls to
repack.repack_indexdef with the third arg being NULL.

Output gets an additional "TABLESPACE pg_default". So add
another variant of expected output to cover 9.6.2, 9.5.6,
9.4.11 and 9.3.16.

Stable update at https://www.postgresql.org/about/news/1733/

Signed-off-by: Christian Ehrhardt <christian.ehrhardt@canonical.com>

### DIFF
--- a/regress/expected/tablespace_2.out
+++ b/regress/expected/tablespace_2.out
@@ -4,7 +4,6 @@ SET client_min_messages = warning;
 --
 -- Note: in order to pass this test you must create a tablespace called 'testts'
 --
-CREATE TABLESPACE testts LOCATION '/tmp/pg-repack-tablespace';
 SELECT spcname FROM pg_tablespace WHERE spcname = 'testts';
  spcname 
 ---------
--- a/regress/expected/tablespace_2.out
+++ b/regress/expected/tablespace_2.out
@@ -1,0 +1,235 @@
+SET client_min_messages = warning;
+--
+-- Tablespace features tests
+--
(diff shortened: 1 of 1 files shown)


MasahikoSawada:
Tested this PR but regression test still doesn't pass.

```
--- /home/masahiko/pgsql/source/pg_repack/regress/expected/tablespace_2.out     Mon Feb 20 16:46:41 2017
+++ /home/masahiko/pgsql/source/pg_repack/regress/results/tablespace.out        Mon Feb 20 16:46:45 2017
@@ -4,7 +4,6 @@
 --
 -- Note: in order to pass this test you must create a tablespace called 'testts'
 --
-CREATE TABLESPACE testts LOCATION '/tmp/pg-repack-tablespace';
 SELECT spcname FROM pg_tablespace WHERE spcname = 'testts';
  spcname
 ---------
```

cpaelzer:
Yeah good catch, in Debian&Ubuntu we have other Delta which makes it into
the expected result as well and as I just copied it over it fails for you
now.
Modifying and updating the PR in just a bit.
​

cpaelzer:
Should be better now, once more I beg a pardon.
Please have a look again if it works for you now.

MasahikoSawada:
Thanks! LGTM :)

schmiddy:
Thanks for the fix!

PR #117, opened 2017-03-22
Support table inheritance

funbringer:
Hi again,

I'd like to introduce a new option: **--parent-table** (**-I**). It'd be useful if we had to repack both partitioned table and its partitions (table inheritance, pg_pathman, pg_partman etc). Child tables are fetched using the new function `repack.get_table_and_inheritors(regclass)`.

I've changed functions `repack_one_database()` and `repack_all_indexes()` in order to achieve the desired behavior:

```bash
# example 1
./bin/pg_repack -Itest_hash debug
INFO: repacking table "test_hash"
INFO: repacking table "test_hash_0"
INFO: repacking table "test_hash_1"
INFO: repacking table "test_hash_2"
INFO: repacking table "test_hash_3"

# example 2
./bin/pg_repack -Itest_hash debug -x
INFO: repacking indexes of "public.test_hash"
INFO: repacking index "public"."test_hash_val_idx"
INFO: repacking indexes of "public.test_hash_0"
INFO: repacking index "public"."test_hash_0_val_idx"
INFO: repacking indexes of "public.test_hash_1"
INFO: repacking index "public"."test_hash_1_val_idx"
INFO: repacking indexes of "public.test_hash_2"
INFO: repacking index "public"."test_hash_2_val_idx"
INFO: repacking indexes of "public.test_hash_3"
INFO: repacking index "public"."test_hash_3_val_idx"

# example 3
./bin/pg_repack -Itest_hash -Itest_range --all debug
ERROR: cannot repack specific table(s) in all databases
```

Feedback is very welcome!


funbringer:
@MasahikoSawada, what do you think of my proposal?

MasahikoSawada:
Thank you for proposing! This feature is very useful for the system using table partitioning.

A few questions,
* When `-Z` option is not specified, should we analyze only parent table rather?
* I don't know the detail of declarative partitioning much which is introduced to PostgreSQL 10 but we will still be able to use this logic (i.g, searching child tables using table inheritance functionality)?

funbringer:
to @MasahikoSawada:

> When -Z option is not specified, should we analyze only parent table rather?

IMHO we should analyze all tables, and it works this way:

```bash
INFO: repacking table "test_range_3_2"
DEBUG: ---- repack_one_table ----
DEBUG: target_name    : test_range_3_2
DEBUG: target_oid     : 24819
DEBUG: target_toast   : 0
DEBUG: target_tidx    : 0
DEBUG: pkid           : 24822
DEBUG: ckid           : 0
DEBUG: create_pktype  : CREATE TYPE repack.pk_24819 AS (val integer)
DEBUG: create_log     : CREATE TABLE repack.log_24819 (id bigserial PRIMARY KEY, pk repack.pk_24819, row test_range_3_2)
DEBUG: create_trigger : CREATE TRIGGER repack_trigger AFTER INSERT OR DELETE OR UPDATE ON test_range_3_2 FOR EACH ROW EXECUTE PROCEDURE repack.repack_trigger('INSERT INTO repack.log_24819(pk, row) VALUES( CASE WHEN $1 IS NULL THEN NULL ELSE (ROW($1.val)::repack.pk_24819) END, $2)')
DEBUG: enable_trigger : ALTER TABLE test_range_3_2 ENABLE ALWAYS TRIGGER repack_trigger
DEBUG: create_table   : CREATE TABLE repack.table_24819 WITH (oids=false) TABLESPACE pg_default AS SELECT val FROM ONLY test_range_3_2
DEBUG: drop_columns   : (skipped)
DEBUG: delete_log     : DELETE FROM repack.log_24819
DEBUG: lock_table     : LOCK TABLE test_range_3_2 IN ACCESS EXCLUSIVE MODE
DEBUG: sql_peek       : SELECT * FROM repack.log_24819 ORDER BY id LIMIT $1
DEBUG: sql_insert     : INSERT INTO repack.table_24819 VALUES ($1.*)
DEBUG: sql_delete     : DELETE FROM repack.table_24819 WHERE (val) = ($1.val)
DEBUG: sql_update     : UPDATE repack.table_24819 SET (val) = ($2.val) WHERE (val) = ($1.val)
DEBUG: sql_pop        : DELETE FROM repack.log_24819 WHERE id IN (
DEBUG: ---- setup ----
DEBUG: index[0].target_oid      : 24822
DEBUG: index[0].create_index    : CREATE UNIQUE INDEX index_24822 ON repack.table_24819 USING btree (val) TABLESPACE pg_default
DEBUG: LOCK TABLE test_range_3_2 IN ACCESS SHARE MODE
DEBUG: No competing DDL to cancel.
DEBUG: Waiting on ACCESS SHARE lock...
DEBUG: ---- copy tuples ----
DEBUG: No competing DDL to cancel.
DEBUG: ---- create indexes ----
DEBUG: Have 1 indexes and num_workers=0
DEBUG: set up index_jobs [0]
DEBUG: target_oid   : 24822
DEBUG: create_index : CREATE UNIQUE INDEX index_24822 ON repack.table_24819 USING btree (val) TABLESPACE pg_default
DEBUG: ---- swap ----
DEBUG: ---- drop ----
DEBUG: ---- analyze ----
```

> I don't know the detail of declarative partitioning much which is introduced to PostgreSQL 10 but we will still be able to use this logic (i.g, searching child tables using table inheritance functionality)?

Precisely! Here's an example:

```plpgsql
/* parent */
create table test_range (val int) partition by range(val);

create table test_range_1 partition of test_range for values from (0) to (100);
create unique index on test_range_1(val);
create table test_range_2 partition of test_range for values from (100) to (200);
create unique index on test_range_2(val);

/* sub-parent */
create table test_range_3 partition of test_range
for values from (200) to (300)
partition by range(val);

create table test_range_3_1 partition of test_range_3 for values from (200) to (250);
create unique index on test_range_3_1(val);
create table test_range_3_2 partition of test_range_3 for values from (250) to (300);
create unique index on test_range_3_2(val);
```

```bash
./bin/pg_repack -I test_range debug
INFO: repacking table "test_range_1"
INFO: repacking table "test_range_2"
INFO: repacking table "test_range_3_1"
INFO: repacking table "test_range_3_2"
```

I just had to patch a few things here and there, like usage of `CatalogIndexInsert()` (it's **static** now), and it seems to work properly.

funbringer:
Merged `master`, all tests should work now.

dvarrazzo:
@MasahikoSawada do you think we should merge this in 1.4?

funbringer:
to @dvarrazzo, @MasahikoSawada:

I've resolved all conflicts just in case.

dvarrazzo:
Thank you: I was trying to do that but couldn't push to your branch; saw you adding patches to resolve conflict and let you do. If @MasahikoSawada is ok with merging this he can do that after adding japanese docs.

bin/pg_repack.c
```diff
@@ -2073,14 +2113,47 @@ repack_all_indexes(char *errbuf, size_t errsize)
...
+
+			if (PQresultStatus(res) != PGRES_TUPLES_OK)
+			{
+				elog(WARNING, "%s", PQerrorMessage(connection));
+				continue;
+			}
+
+			nchildren = PQntuples(res);
+
+			if (nchildren == 0)
```
MasahikoSawada: I guess nchildren is never 0. Is it better to use `assert` instead?

funbringer: Perhaps, but it might trigger in case of concurrent table drop. I don't take any locks since we use the resulting set of Oids in WHERE clause.

MasahikoSawada: > I don't take any locks since we use the resulting set of Oids in WHERE clause.

Why? Can't we take even AccessShareLock? I guess it's better to prevent from the current table drop.

funbringer: @MasahikoSawada, ok, I'll make the adjustments.

MasahikoSawada: Thanks!

funbringer: @MasahikoSawada, actually, I've inspected `regclassin` and it seems that casts to `regclass` type don't take any locks on relations.

As far as I understand, pg_repack doesn't take locks in `repack_one_database()`. Here's a test:

Step 1:
```bash
# terminal A
gdb --args ./bin/pg_repack -t test db
...
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./bin/pg_repack...done.

(gdb) b repack_one_table 
Breakpoint 1 at 0x404828: file pg_repack.c, line 1101.

gdb) r
Starting program: /home/dmitry/Projects/pg_repack/bin/pg_repack -t test db
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".

# ok, struct 'table' should be filled with Oids etc
Breakpoint 1, repack_one_table (table=0x7fffffffdc80, orderby=0x0) at pg_repack.c:1101
1101            PGresult           *res = NULL
```

Step 2:
```plpgsql
-- terminal B

 -- no lock on table 'test'
select locktype, relation::regclass, mode from pg_locks ;
  locktype  | relation |      mode       
------------+----------+-----------------
 relation   | pg_locks | AccessShareLock
 virtualxid |          | ExclusiveLock
(2 rows)

 -- we can drop it
drop table test;
DROP TABLE
```

Step 3:
```bash
# terminal A
(gdb) c
Continuing.
INFO: repacking table "test"
ERROR:  relation "test" does not exist
DETAIL: query was: RESET statement_timeout
ERROR: query failed: ERROR:  table name not found for OID 1339794
DETAIL: query was: SELECT repack.repack_drop($1, $2)
ERROR: query failed: ERROR:  table name not found for OID 1339794
DETAIL: query was: SELECT repack.repack_drop($1, $2)
... (loop)
```

I'm sure that this problem should be addressed. Concurrent `DROP TABLE` is possible, so I decided to check the number of result tuples at least in code I've changed in my patch.

We should add a few checks here and there (probably in `repack_one_table()` too), or take locks beforehand. I didn't dare to make such changes.

funbringer: @MasahikoSawada,

Totally forgot to describe possible changes to `repack_get_table_and_inheritors()`.

We could, for example, take a lock on parent and traverse children tree with `AccessShareLock`:

```C
Datum
repack_get_table_and_inheritors(PG_FUNCTION_ARGS)
{
	Oid	    parent = PG_GETARG_OID(0);
	List	   *relations;
	Datum	   *relations_array;
	int	    relations_array_size;
	ArrayType  *result;
	ListCell   *lc;
	int         i;

        /* take locks HERE! */
        LockRelationOid(parent, AccessShareLock);
	relations = find_all_inheritors(parent, AccessShareLock, NULL);

	relations_array_size = list_length(relations);
	if (relations_array_size == 0)
		PG_RETURN_ARRAYTYPE_P(construct_empty_array(OIDOID));
```

But the problem remains: the `parent` table might be dead when we finally get to the ` LockRelationOid(parent, AccessShareLock);` since `'TABLE_NAME'::regclass` won't lock it.

MasahikoSawada: Since the transaction executing repack_get_table_and_inheritors closes at end of repack_one_database function, the table could be dropped before repacking it actually.
What I'm concerned is such SQL executing repack_get_table_and_inheritors can return a inconsistent result. Maybe I worry too  much, though.

funbringer: @MasahikoSawada, I think we might see some dropped child tables or the tables that are not children anymore, but that's ok since these tables might change after `repack_one_database()` finishes even if we take `AccessShareLock`.

My thought was that either we should not bother at all, or we should at least take some precautions in `repack_one_table()` and other functions too.

If you still insist on taking locks in `repack_get_table_and_inheritors()`, I'll make the appropriate changes.

P.S. Forgot to thank you for the review! :)

MasahikoSawada: IMHO, just to be on the safe side I'd like to acquire AccessShareLock in that function.

funbringer: @MasahikoSawada, done! Everything seems to work properly.

lib/repack.c
```diff
@@ -1289,3 +1294,52 @@ repack_index_swap(PG_FUNCTION_ARGS)
...
+	Datum	   *relations_array;
+	int			relations_array_size;
+	ArrayType  *result;
+	ListCell   *lc;
+	int			i;
+
+	LockRelationOid(parent, AccessShareLock);
+
+	/* Check that parent table exists */
+	if (!SearchSysCacheExists1(RELOID, ObjectIdGetDatum(parent)))
```
MasahikoSawada: Why do we need to check here? Is it possible that the relation is dropped after we got AccessShareLock?

lib/repack.c
```diff
@@ -1289,3 +1294,52 @@ repack_index_swap(PG_FUNCTION_ARGS)
...
+		PG_RETURN_ARRAYTYPE_P(construct_empty_array(OIDOID));
+
+	/* Also check that children exist */
+	relations = find_all_inheritors(parent, AccessShareLock, NULL);
```
MasahikoSawada: Well, I think we should release these locks after used. If not, we should add the comment to this function instead.

MasahikoSawada:
> @MasahikoSawada do you think we should merge this in 1.4?

@dvarrazzo Thank you for looking at this item! Almost 1 month passed since when we started the release work, I guess there are people who wait for new version. I want to merge this feature in v1.4, but because this is a not bug fix, If this patch could not get the stable state by end of this month I think we can push it out to the next version. That is, we release new version at least by 30th April. Thought?

lib/repack.c
funbringer: > Is it possible that the relation is dropped after we got AccessShareLock, or we get AccessShareLock on a dropped relation?

There's a possibility that we get a lock on a dropped relation, since this function will be called on a string literal cast to `regclass` type (`'table_name'::regclass`), here:

```plpgsql
relid = ANY(repack.get_table_and_inheritors($%d::regclass))
```
and also here:
```plpgsql
WHERE c.oid = ANY (repack.get_table_and_inheritors($1::regclass))
```

As I've already mentioned, `regclassin()` (the implementation of `::regclass` cast) doesn't take any locks:

```C
/* We might not even have permissions on this relation; don't lock it. */
result = RangeVarGetRelid(makeRangeVarFromNameList(names), NoLock, false);
```

IMHO It's better to exit early if we find out that relation doesn't exist anymore.

lib/repack.c
funbringer: @MasahikoSawada, why? They'll be released automatically after transaction finishes. We're not going to benefit from releasing the locks right after we took them.

Should I add a comment above the function's definition?

lib/repack.c
MasahikoSawada: Okay, I understood. 

> IMHO It's better to exit early if we find out that relation doesn't exist anymore.

Seems good idea.

lib/repack.c
MasahikoSawada: I guess that you're assuming this function is called only in a SQL command. I thought that we use this function in other C function in the future. In this case it's better to release by end of function or document it.

dvarrazzo:
@MasahikoSawada I'm ok to set the 30th of April as a deadline for release. If we don't have this feature in by that date we can ship anyway.

Thank you immensely for your work!

lib/repack.c
funbringer: > I guess that you're assuming this function is called only in a SQL command.

Mostly, yes. That might change, of course.

I decided to extend a `@brief` section of repack_get_table_and_inheritors().

MasahikoSawada: Thanks!

MasahikoSawada:
Merged.
@funbringer I appreciate all your efforts!

funbringer:
thanks!

PR #118, opened 2017-03-23
Update pg_repack_jp.rst

MasahikoSawada:
Since v1.3.3, when Japanese doc has been introduced, this
file had not been maintained.


MasahikoSawada:
Because this fixes only Japanese doc, I will merge it.

Issue #10, opened 2013-07-17
storage parameter on TOAST table gets lost


pierrestroh:
hi.

when using pg_repack/pg_reorg (same result) i recognized a missing storage parameter on my TOAST table after reorganization.

the table had set toast.autovacuum_vacuum_scale_factor to 0.1 (via ALTER TABLE mytable SET (toast.autovacuum_vacuum_scale_factor=0.1); )
when running pg_reorg/pg_repack on such table, the storage parameter for the TOAST table was gone.

used command was: pg_repack -n -e -U postgres -d playground -t mytable

server-version: PostgreSQL 9.1.6
pg_repack version 1.1.8

dvarrazzo:
Tentatively work on this issue again. The storage parameters are available on three objects of our interest:

- the tables (in `pg_class.reloptions`)
- the single columns (in `pg_attribute.attoptions`)
- the indexes (in `pg_class.reloptions` too)

there is a remote chance that, because of the way we rebuild indexes, they end up being correct, but if this is the case it has to be tested and added to the test suite anyway.

The tables and the columns require options restoring. This should be done after table creation and before loading data (so e.g. `fillfactor` is honoured etc.)

dvarrazzo:
Note, there is adequate testing that indexes reloptions are handled ok

dvarrazzo:
(I know that's not what the OP requested, but): added test to verify the columns options are respected too. Merging #123 should merge this bug.

PR #123, opened 2017-04-03
Specify the storage option for TOAST table when create table.

MasahikoSawada:
The storage option such as AUTOVACUUM_VACUUM_SCALE_FACTOR can be
set to both heap table and TOAST table. But the storage parameter
for TOAST table had gone after repacked. This change create new
function get_storage_param which returns all storage paramters
including for TOAST table and OID setting.

Issue #10.


MasahikoSawada:
There are some options to fix this issue. I've selected a option to make pg_repack specify the all storage parameter including parameters for TOAST table and OID setting when creating table. Please give me feedback.

dvarrazzo:
It seems a good approach factoring out the parameters list into a function of its own. I'll sort out this conflict I've made in master.

dvarrazzo:
I have rebased this branch to simplify history (dropping merges into master to solve conflicts in tests)

MasahikoSawada:
Thank you for reviewing! Barring any objections, I'll merge this tomorrow.

PR #124, opened 2017-04-10
Drop old versions

dvarrazzo:
This MR drops support for PG version before the introduction of CREATE EXTENSION, simplifying largely C code, build and testing. @MasahikoSawada @schmiddy are you ok with it?


MasahikoSawada:
+1!

PR #131, opened 2017-06-03
Add "-X" option to avoid waiting for other transactions

farrokhi:
There was a conservative wait[1] for all existing transaction to
finish before repacking a table. This may not be desired on a busy
database on which you have long running transaction, which may block
pg_repack to do the job. This option will display a warning message,
but continues the process in such cases.

[1] http://lists.pgfoundry.org/pipermail/reorg-general/2015-October/000318.html

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -239,6 +239,7 @@ static bool sqlstate_equals(PGresult *res, const char *state)
 
 static bool				analyze = true;
 static bool				alldb = false;
+static bool				no_paranoia = false;
 static bool				noorder = false;
 static SimpleStringList	parent_table_list = {NULL, NULL};
 static SimpleStringList	table_list = {NULL, NULL};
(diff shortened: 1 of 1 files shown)


andreasscherbaum:
@farrokhi Can you please rebase the PR? This will also trigger the checks.
An update for the documentation is also missing.

The name "paranoia" is maybe not the best choice, can you think of a better name?

farrokhi:
Thanks for the feedback. I updated the PR (including documents). Please kindly review.

andreasscherbaum:
Did you rebase the PR? It includes unrelated changes from other commits.

farrokhi:
I cleaned up the same PR. It should have a single commit with relevant changes now.

andreasscherbaum:
File-wise it looks good. Will place it in our next review cycle. Thanks.

dvarrazzo:
Why `-X` at all?

I have looked at them, and Every Single Option has a short option. It seems a bad idea (thankfully we have almost run over the entire alphabet).

I would start from here to stop adding one-letter options.

Maybe we should also deprecate most of them, except the most obvious ones, in a future non-backward compatible version.

PR #132, opened 2017-06-04
Avoid return from lock_exclusive() while transaction aborted ..

justinpryzby:
..causing "ERROR: DROP INDEX CONCURRENTLY cannot run inside a transaction block"
github#129


justinpryzby:
need to fix commit message..

PR #133, opened 2017-06-04
Avoid return from lock_exclusive() while in transaction ..

justinpryzby:
..causing "ERROR: DROP INDEX CONCURRENTLY cannot run inside a transaction block"
github#129

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -1771,6 +1771,7 @@ lock_exclusive(PGconn *conn, const char *relid, const char *lock_query, bool sta
 			{
 				elog(WARNING, "timed out, do not cancel conflicting backends");
 				ret = false;
+				pgut_rollback(conn);
 				break;
 			}
 			else


MasahikoSawada:
Thank you for the PR.

Yeah, it seems a bug of `--no-kill-backend` option. I think that it's not good idea to do `pgut_rollback()` at that point in `lock_exclusive()`, because `repack_one_table()` does `pgut_rollback()` after `lock_exclusive()` failed and if `start_xac` is true we should do `ROLLBACK TO SAVEPOINT repack_sp1` instead. How about doing `pgut_rollback()` after `lock_exclusive()` failed in `repack_table_indexes()`?

MasahikoSawada:
After more thought, I found we should rollback in lock_exclusive as you mentioned. But it also need to support if start_xact = false. I've opened a new PR #143, please review it.
Sorry, I'm not familiar with the use of github so maybe it's not good a way to open an another  PR for the same purpose.

dvarrazzo:
This has been merged as #143.

PR #136, opened 2017-07-12
Fix a small memory leak in repack_one_database function.

MasahikoSawada:
Total 3 issues were noticed by the coverity scan but other two issues were not problem actually.


MasahikoSawada:
Will merge it, barring any objections.

MasahikoSawada:
Thank you! Merged.

Issue #138, opened 2017-08-01
pg_repack not "organizing" table.


acmzero:
I'm trying to organize a table based on a timestamp column, I executed -o "start_time" and I see everything finishes successfully but the rows inside the pages are not ordered at all.

After digging deeper I found that when copying the rows from the original table to the temporal table there is no `ORDER BY` clause, leaving the table in the same order as the original.

I manage to order the table by modifying the copy_data column at view repack.tables  by appending 
`|| (case when repack.oid2text(r.oid)='target_table' then ' order by start_time' else '' end)` 

You can find more details here:
[pg_repack not organizing table](https://dba.stackexchange.com/questions/182315/pg-repack-not-organizing-table-by-column)

MasahikoSawada:
Thank you for reporting.

I could reproduce this issue, and it seems to me that it's a bug of pg_repack. commit 5adff6ff0b88d6f162719eff7176069730537c2a added this bug. We should have add `ORDER BY` to the copy SQL. I'll fix it on this weekend.

dvarrazzo:
Fixed in pg_repack 1.4.1, just released.

acmzero:
I can confirm this is fixed.

Awesome!!!

Thanks.

PR #140, opened 2017-08-04
Fix reorganize table without sorting.

MasahikoSawada:
Commit 5adff6ff0b88d6f162719eff7176069730537c2a separated the
data copy from creating table. This is a cause of bug that
pg_repack doesn't actually sort table during reorganization.
This commit fixes this issue by adding ORDER BY clause to Copy
SQL rather than CREATE TABLE SQL.

Reported by acmzero on issue #138.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -821,7 +821,7 @@ repack_one_database(const char *orderby, char *errbuf, size_t errsize)
 		appendStringInfoString(&sql, tablespace);
 		appendStringInfoString(&sql, create_table_2);
 
-		/* Always append WITH NOT DATA to CREATE TABLE SQL*/
+		/* Always append WITH NO DATA to CREATE TABLE SQL*/
 		appendStringInfoString(&sql, " WITH NO DATA");
 		table.create_table = sql.data;
 
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -772,6 +772,7 @@ repack_one_database(const char *orderby, char *errbuf, size_t errsize)
 	for (i = 0; i < num; i++)
 	{
 		repack_table	table;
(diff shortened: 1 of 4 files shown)


bin/pg_repack.c
```diff
@@ -814,17 +815,27 @@ repack_one_database(const char *orderby, char *errbuf, size_t errsize)
 		table.sql_pop = getstr(res, i, c++);
 		tablespace = getstr(res, i, c++);
 
+		/* Craft CREATE TABLE SQL */
 		resetStringInfo(&sql);
 		appendStringInfoString(&sql, create_table_1);
 		appendStringInfoString(&sql, tablespace);
 		appendStringInfoString(&sql, create_table_2);
+
+		/* Always append WITH NOT DATA to CREATE TABLE SQL*/
```
cbandy: Should match `WITH NO DATA` on the following line?

MasahikoSawada: Yes, fixed.

MasahikoSawada:
Barring any objections, I'll merge this PR.

dvarrazzo:
Sure, thank you very much! :+1:

sean-:
Please.  I was looking at this code last week and ramping up to fix it.  Not having `-o` work anymore was a rather unpleasant surprise.

MasahikoSawada:
Merged. Since it's an relatively serious issue should we release a new version including this fix?

dvarrazzo:
@MasahikoSawada I will make a release today. Is it possible for you to fix the Japanese readme to add a note? However I will not delay the release for this, we can add it later no problem ;)

Thank you very much for your timely work on this problem!

MasahikoSawada:
Thank you for prompt response and releasing new version!
I've updated the Japanese readme.

PR #147, opened 2017-08-28
Add CLI argument for "max tuples before table swap"

### LINKED ISSUE #126: CLI option for MIN_TUPLES_BEFORE_SWITCH
I've been running into an issue recently where pg_repack gets stuck replaying the log before swapping tables. Our query volume on that table seems to be high enough that the hardcoded value of `MIN_TUPLES_BEFORE_SWITCH 20` is not enough.

It seems to me that the trade-off with a higher minimum value is a longer time spent holding the access exclusive lock during swap. Is there anything else I'm not thinking of?

I'd like to propose adding a CLI option for specifying that value. Doing so would allow even the busiest tables to be swapped, and the trade-off with lock time is entirely in the hands of the user.

If there is interest here, I'd be happy to put together a patch.

P.S. Thanks for continuing to support this incredibly useful tool. :smile:

jwilm:
This adds a new argument

```
  -m, --max-swap-tuples=NUM   max tuples allowed in replay log at table swap
```

Resolves #126.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -48,11 +48,6 @@ const char *PROGRAM_VERSION = "unknown";
  */
 #define APPLY_COUNT		1000
 
-/* Once we get down to seeing fewer than this many tuples in the
- * log table, we'll say that we're ready to perform the switch.
- */
-#define MIN_TUPLES_BEFORE_SWITCH	20
-
 /* poll() or select() timeout, in seconds */
 #define POLL_TIMEOUT    3
 
@@ -254,6 +249,7 @@ static bool				dryrun = false;
 static unsigned int		temp_obj_num = 0; /* temporary objects counter */
 static bool				no_kill_backend = false; /* abandon when timed-out */
 static bool				no_superuser_check = false;
+static unsigned int		max_tuples_before_switch = 20;
 static SimpleStringList	exclude_extension_list = {NULL, NULL}; /* don't repack tables of these extensions */
 
 /* buffer should have at least 11 bytes */
@@ -284,6 +280,7 @@ static pgut_option options[] =
 	{ 'b', 'D', "no-kill-backend", &no_kill_backend },
 	{ 'b', 'k', "no-superuser-check", &no_superuser_check },
 	{ 'l', 'C', "exclude-extension", &exclude_extension_list },
+	{ 'u', 'm', "max-swap-tuples", &max_tuples_before_switch },
 	{ 0 },
 };
 
@@ -1436,12 +1433,12 @@ repack_one_table(repack_table *table, const char *orderby)
 		/* We'll keep applying tuples from the log table in batches
 		 * of APPLY_COUNT, until applying a batch of tuples
 		 * (via LIMIT) results in our having applied
-		 * MIN_TUPLES_BEFORE_SWITCH or fewer tuples. We don't want to
+		 * max_tuples_before_switch or fewer tuples. We don't want to
 		 * get stuck repetitively applying some small number of tuples
 		 * from the log table as inserts/updates/deletes may be
 		 * constantly coming into the original table.
 		 */
-		if (num > MIN_TUPLES_BEFORE_SWITCH)
+		if (num > max_tuples_before_switch)
 			continue;	/* there might be still some tuples, repeat. */
 
 		/* old transactions still alive ? */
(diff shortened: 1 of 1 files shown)


MasahikoSawada:
Thank you for the PR!
I think documentation should be updated.

dvarrazzo:
@MasahikoSawada are you ok to merge this MR if the documentation is completed?

MasahikoSawada:
Yes. I think this option would be useful for users and sometimes I got the similar requests from our clients. The code is simple and looks good to me.

dvarrazzo:
It's not polite to release a new command line switch in a bugfix release, is it? I guess we should include it in a new "major release"

cbandy:
When the behavior of the same call (no `-m`) stays the same, it can go out in a patch imo.

If, later, one decides to change the default of the arg, that would be breaking and major.

MasahikoSawada:
I think a new command should be added to a new major version, too. So we can either bump up versions to 1.5 that includes the bugfix and the new command, or release version 1.4.3 including the bugfix and release version 1.5 including both of them.

schmiddy:
I think it's good to make this knob configurable.

One question though, why the proposed renaming from `MIN_TUPLES_BEFORE_SWITCH` to `max_tuples_before_switch`. The threshold here really is a "minimum" -- once we detect there are fewer than X number of rows left in the repack table, we perform the swap.

I can't wrap my head around how this could be a "max", since the repack table will start with a large number of rows and work its way down. The current logic of the code is:
```
		if (num > MIN_TUPLES_BEFORE_SWITCH)
 			continue;	/* there might be still some tuples, repeat. */
```
which makes sense to me -- if we are above a minimum, ignore and keep going. Saying "if we are **above a maximum**, ignore and keep going", doesn't quite make sense to me.

jwilm:
@schmiddy My apologies for the confusion here. Please allow me to explain why I renamed it.

When I was adding the argument, the "minimum" terminology started to confuse me. From a user's perspective, you want to limit how many rows need to be moved at table swap time. More rows = more time holding a lock. It can be viewed as a maximum of how many rows you are willing to process while holding that lock.

For example, If we said "I want a minimum of 1,000 rows in a table before swapping", it seems that any number greater than 1,000 would suffice, but really a number less than that is desired. "I want a **maximum** of 1,000  rows at swap time" makes it clear that there will be no more than 1,000.

This argument seems good for the CLI argument name. As for the code, I figured it would be better to keep the internal naming similar to the CLI argument to avoid confusion.

bin/pg_repack.c
```diff
@@ -284,6 +280,7 @@ static pgut_option options[] =
 	{ 'b', 'D', "no-kill-backend", &no_kill_backend },
 	{ 'b', 'k', "no-superuser-check", &no_superuser_check },
 	{ 'l', 'C', "exclude-extension", &exclude_extension_list },
+	{ 'u', 'm', "max-swap-tuples", &max_tuples_before_switch },
```
TysonAndre, line 283: There should be a sanity check that `max_tuples_before_switch < APPLY_COUNT`

jwilm: The check in the code is against the number of tuples remaining in the log. Wouldn't your statement only be true if that number was < APPLY_COUNT + max_tuples_before_switch? I don't see how this change is going to result in early termination of that loop.

TysonAndre:
```
		/* We'll keep applying tuples from the log table in batches
		 * of APPLY_COUNT, until applying a batch of tuples
		 * (via LIMIT) results in our having applied
		 * MIN_TUPLES_BEFORE_SWITCH or fewer tuples. We don't want to
		 * get stuck repetitively applying some small number of tuples
		 * from the log table as inserts/updates/deletes may be
		 * constantly coming into the original table.
		 */
```

No, it's the number of tuples that were fetched in that batch (identical to the number that were applied), not the total number of rows remaining in the log. Doing a full count would be slower, so repack_apply wouldn't compute that

Also see lib/repack.c

```c
 * @retval				Number of performed operations.
 */
Datum
repack_apply(PG_FUNCTION_ARGS)
```

TysonAndre:
What seems to need to be changed in the docs:

The new output of `pg_repack --help` with the new indentation, if needed should be replaced in `doc/pg_repack.rst` This should be added to the reorg options section of pg_repack.rst.

It seems like the release notes have been put together separately from the feature PRs in the past?

TysonAndre:
Other feature ideas I had for solving related issues, out of the scope of this PR.

- By default, count the number of times that the applied count was less than the batch size. Once the total number of times this happens for a table reaches 100 or so, assume that repack is somehow misconfigured (extremely high rate of inserts, client network round trip time is too high, or some combination of issues) and just initialize the final switch.

  Also, I'm not that familiar with repack or postgres, but is there a reason why pg_repack can't have an option to apply the remaining rows to repack.table_$id after `lock original_table_name`?
- Maybe set up a signal handler. For example, if the repack client recieves SIGUSR1, then raise the repack limit to `APPLY_COUNT - 1` so that an administrator can ensure a running misconfigured repack will finish. There would be less broadly useful if this checked for the total number of incomplete batches.

TysonAndre:
> For example, If we said "I want a minimum of 1,000 rows in a table before swapping", it seems that any number greater than 1,000 would suffice, but really a number less than that is desired. "I want a maximum of 1,000 rows at swap time" makes it clear that there will be no more than 1,000.
> 
> This argument seems good for the CLI argument name. As for the code, I figured it would be better to keep the internal naming similar to the CLI argument to avoid confusion.

Maybe the name `tuples_applied_in_batch_threshold_before_switch` or something would avoid confusion, but all of the names I can think of are confusing since this mode of operation doesn't have a convenient way to refer to it

bin/pg_repack.c
```diff
@@ -2222,21 +2219,22 @@ pgut_help(bool details)
...
+	printf("  -N, --dry-run               print what would have been repacked\n");
+	printf("  -j, --jobs=NUM              Use this many parallel jobs for each table\n");
+	printf("  -i, --index=INDEX           move only the specified index\n");
+	printf("  -x, --only-indexes          move only indexes of the specified table\n");
+	printf("  -T, --wait-timeout=SECS     timeout to cancel other backends on conflict\n");
+	printf("  -D, --no-kill-backend       don't kill other backends when timed out\n");
+	printf("  -Z, --no-analyze            don't analyze at end\n");
+	printf("  -k, --no-superuser-check    skip superuser checks in client\n");
+	printf("  -C, --exclude-extension     don't repack tables which belong to specific extension\n");
+	printf("  -m, --max-swap-tuples=NUM   max tuples allowed in replay log at table swap\n");
```
TysonAndre, line 2239: `trigger table swap once the last batch replayed by this was of size NUM or less` might be more accurate, but harder to understand.

Mentioning `(default=20)` might help end users understand what changing this would do

andreasscherbaum:
Handled in this PR: https://github.com/reorg/pg_repack/pull/347

PR #149, opened 2017-09-29
Clarify note about GiST indexes

craigds:
As per http://lists.pgfoundry.org/pipermail/reorg-general/2015-May/000306.html


schmiddy:
Thanks!

Issue #130, opened 2017-05-29
Is -E (--encoding) option necessary for initdb? 


kfuru:
Hello,
I tested pg_repack regression in CentOS7.
But, one test FAILED in my environment.

To create PostgreSQL database cluster, I did the following.
 initdb --no-locale  (Encoding is SQL_ASCII) 

<pre> 
============== running regression test queries        ==============
 test init-extension           ... ok
 test repack                   ... FAILED
 test tablespace               ... ok
 test issue3                   ... ok
</pre>

Looking at regression.diffs,
It seems that it error with INSERT of test data.

<pre>
 $ cat ./regress/regression.diffs
 *** /var/lib/pgsql/pg_repack-ver_1.4.0.1/regress/expected/repack.out    2017-04-26 01:58:09.000000000 +0900
 --- /var/lib/pgsql/pg_repack-ver_1.4.0.1/regress/results/repack.out     2017-05-23 09:49:00.420905000 +0900
 ***************
 *** 98,103 ****
 --- 98,104 ----
   INSERT INTO tbl_badindex VALUES(2, 10);
   -- insert data that is always stored into the toast table if column type is extended.
   INSERT INTO tbl_with_mod_column_storage SELECT 1, array_to_string(ARRAY(SELECT chr(code) FROM generate_series(33,3000) code), '');
 + ERROR:  requested character too large for encoding: 256
   -- This will fail. Silence the message as it's different across PG versions.
   SET client_min_messages = fatal;
   CREATE UNIQUE INDEX CONCURRENTLY idx_badindex_n ON tbl_badindex (n);
 
 ======================================================================
</pre>

Is it mandatory to specify the character code using the -E/--encoding option (e.g.-E UTF8) in this test?

Regards,
kfuru

MasahikoSawada:
As commented in the test, we here need a long string value which is stored into a toast table. Should we change the regression test so that it generates a long random string using only characters that are valid in SQL-ASCII encording? Or should we leave and mention this as a restriction in the docs?

dvarrazzo:
Do we only care about the string being long enough? If so we can limit to just printable ascii and avoid any encoding problem. What about:

    SELECT 1, array_to_string(ARRAY(SELECT chr(code % (127 - 32) + 32)
    FROM generate_series(1,3000) code), '')

dvarrazzo:
Fixing this

MasahikoSawada:
[...]

PR #150, opened 2017-10-03
Change to use only the characters that are printable ASCII.

MasahikoSawada:
The regression test could fail on some environments such as where
the encoding is SQL_ASCII.

Fixed issue #130.


MasahikoSawada:
Barring any objections, I'll merge it tomorrow.

MasahikoSawada:
Thanks @schmiddy , merged.

Issue #152, opened 2017-10-07
Doubts about repacked tables relfrozenxid


dvarrazzo:
I'm down in the rabbit hole trying to understand #135. Looking at what the `swap_head_or_index_files()` does [here](https://github.com/reorg/pg_repack/blob/ver_1.4.1/lib/repack.c#L1131), it seems that the original author wanted to set the repacked table's relfrozenxid to the greatest of the two, but the code does the opposite (or better it sets the relfrozenxid on the temporary table that then it drops).

```
contrib_regression=# select  relfrozenxid, txid_current() from pg_class where oid = 'test'::regclass;
 relfrozenxid | txid_current 
--------------+--------------
       444833 |       444901
(1 row)

contrib_regression=# vacuum full test;
VACUUM
contrib_regression=# select  relfrozenxid, txid_current() from pg_class where oid = 'test'::regclass;
 relfrozenxid | txid_current 
--------------+--------------
       444902 |       444903
(1 row)

contrib_regression=# cluster test using test_pkey;
CLUSTER
contrib_regression=# select  relfrozenxid, txid_current() from pg_class where oid = 'test'::regclass;
 relfrozenxid | txid_current 
--------------+--------------
       444905 |       444906
(1 row)
```

What `pg_repack` does is sort of backwards:

```
contrib_regression=# \! /usr/lib/postgresql/9.6/bin/pg_repack -d contrib_regression -t test
INFO: repacking table "test"

# adding a couple of LOG in the code:
LOG:  relform1->relfrozenxid = 444905
LOG:  relform2->relfrozenxid = 444908

contrib_regression=# select  relfrozenxid, txid_current() from pg_class where oid = 'test'::regclass;
 relfrozenxid | txid_current 
--------------+--------------
       444905 |       444917
(1 row)
```

shouldn't we set the repacked table frozen xid to the one of the new table?

schmiddy:
You may already know all this @dvarrazzo , but just to summarize some code archaelogy from a little digging I did: the comment above our `swap_heap_or_index_files ` [says](https://github.com/reorg/pg_repack/blob/4b86bb1c4117e56d2f69418dccfb7b46892c29fd/lib/repack.c#L1083-L1084):

> This is a copy of swap_relation_files in cluster.c, but it also swaps relfrozenxid.

[...]

PR #157, opened 2017-11-01
Fix relfrozenxid, relminmxid and relallvisible on table data files swap

bashtanov:
Currently pg_repack sets both `relfrozenxid`s to the older one (the code comment says "larger", but `TransactionIdFollows` acts as `>`, so it essentially works the opposite). It implies excessive anti-wraparound vacuums on tables that have been repacked recently. It seems correct to me to swap this attribute along with the data, regardless whether the xid is normal or not. Same for relminmxid, and for relallvisible, which is a part of statistics.

See also the discussion at https://github.com/reorg/pg_repack/issues/152.

### DIFF
--- a/lib/repack.c
+++ b/lib/repack.c
@@ -1092,6 +1092,7 @@ swap_heap_or_index_files(Oid r1, Oid r2)
 	Form_pg_class relform1,
 				relform2;
 	Oid			swaptemp;
+	TransactionId swaptempxid;
 	CatalogIndexState indstate;
 
 	/* We need writable copies of both pg_class tuples. */
@@ -1128,15 +1129,18 @@ swap_heap_or_index_files(Oid r1, Oid r2)
 	relform1->reltoastrelid = relform2->reltoastrelid;
 	relform2->reltoastrelid = swaptemp;
 
-	/* set rel1's frozen Xid to larger one */
-	if (TransactionIdIsNormal(relform1->relfrozenxid))
-	{
-		if (TransactionIdFollows(relform1->relfrozenxid,
-								 relform2->relfrozenxid))
-			relform1->relfrozenxid = relform2->relfrozenxid;
-		else
-			relform2->relfrozenxid = relform1->relfrozenxid;
-	}
+	/*
+	 * Swap relfrozenxid and relminmxid, as they must be consistent with the data
+	 */
+	swaptemp = relform1->relfrozenxid;
+	relform1->relfrozenxid = relform2->relfrozenxid;
+	relform2->relfrozenxid = swaptempxid;
+
+#if PG_VERSION_NUM >= 90300
+	swaptemp = relform1->relminmxid;
+	relform1->relminmxid = relform2->relminmxid;
+	relform2->relminmxid = swaptempxid;
+#endif
 
 	/* swap size statistics too, since new rel has freshly-updated stats */
 	{
@@ -1154,6 +1158,12 @@ swap_heap_or_index_files(Oid r1, Oid r2)
 		swap_tuples = relform1->reltuples;
 		relform1->reltuples = relform2->reltuples;
 		relform2->reltuples = swap_tuples;
+
+#if PG_VERSION_NUM >= 90200
+		swap_pages = relform1->relallvisible;
+		relform1->relallvisible = relform2->relallvisible;
+		relform2->relallvisible = swap_pages;
+#endif
 	}
 
 	indstate = CatalogOpenIndexes(relRelation);


schmiddy:
@dvarrazzo @MasahikoSawada do you guys have an opinion about this fix?

Thanks for posting this @bashtanov , sorry it is taking us so long to review.

dvarrazzo:
It is probably a problem that must be fix. But the entire swapping has to be fast-forwarded about 10 years. I started working on it at the time @bashtanov sent his MR so it was conflicting with what I was doing too, but then I didn't get around progressing with that work so much.

The behaviour in #152 is described as *bad* (see Robert Haas tranchant response in [the thread](https://www.postgresql.org/message-id/CA%2Bmi_8YWReON2gVk9qoeJLRVzWTSSoLkKAX9DERL36-n4Y8rZg%40mail.gmail.com)) So we should likely merge this. But I still wonder why it was that way in first place.

For me, I'd call it an oversight of the original error and I'd put back the xids the way they were meant to be. This MR also handles `relallvisible`, `relminmxid`, which likely weren't available where the postgres code was cargoculted into pg_reorg. I didn't do my digging into the postgres archaeology to understand if that's right or wrong and I'm deeply disappointed nobody in the core team cares.

xzilla:
I've been meaning to add to this discussion as a few months ago we used repack as part of an xid wraparound avoidance operation, which caused us to look into this a bit deeper than most. 

One of the early things we reasoned was that due to the way repack creates the new table and inserts data, all rows would be written with a current xid within the system. This is fairly evident even from the SQL level and we were able to run a number of checks to confirm this. I should maybe mention that the table in question was over 1TB in size, and had hundreds of millions of rows, and roughly 20 different indexes. 

Once we were happy from the SQL level, we did some relevant investigation of the Postgres source to see if we could find any code areas that we should be concerned about, but again because we had recreated the table and all indexes entirely during the operation, we saw no such cases. Furthermore, new rows and updates rows also behaved as expected, in fact the only piece that seemed missing was pg_class.relfrozenxid which contained the value from the original pre-repacked table. 

At this point we felt confident we could update pg_class.relfrozenxid to match the earliest xid within the data itself and be safe, but before that, we reached out to our upstream cloud provider to have their postgres internals team (yes, we have an usually good relationship with them) also review our findings, and their team concurred it would be safe. Once the system catalog was updated, we saw no issues with performance nor system errors, and a subsequent vacuum verbose and vacuum freeze of the table raised no issues, and we have not seen any further issues having now run for several months. 

So the summary is, consider me a +1 for updating relfrozenxid of the repacked table. As mentioned elsewhere in this discussion, I think that was the original intent of the reorg authors, and at least as of v14 I see no reason why this would cause issues, with the only caveat being I have not looked into the nearby logical replication issue to understand how it might affect that, but in the common case this would be solid improvement for existing repack users, imho.

andreasscherbaum:
@bashtanov Since this PR changes quite a few internal values, is it possible to check that it is working as expected? Other than that, it looks good.

lib/repack.c
```diff
@@ -1128,15 +1129,18 @@ swap_heap_or_index_files(Oid r1, Oid r2)
...
-		if (TransactionIdFollows(relform1->relfrozenxid,
-								 relform2->relfrozenxid))
-			relform1->relfrozenxid = relform2->relfrozenxid;
-		else
-			relform2->relfrozenxid = relform1->relfrozenxid;
-	}
+	/*
+	 * Swap relfrozenxid and relminmxid, as they must be consistent with the data
+	 */
+	swaptemp = relform1->relfrozenxid;
```
bashtanov, line 1135: swaptempxid not swaptemp should be initialized here

lib/repack.c
```diff
@@ -1128,15 +1129,18 @@ swap_heap_or_index_files(Oid r1, Oid r2)
...
+	relform1->relfrozenxid = relform2->relfrozenxid;
+	relform2->relfrozenxid = swaptempxid;
+
+#if PG_VERSION_NUM >= 90300
+	swaptemp = relform1->relminmxid;
```
bashtanov, line 1140: swaptempxid not swaptemp should be initialized here

bashtanov:
I believe it doesn't as it uses an uninitialized variable. I'll test it when I have a moment but it may take similarly long.

za-arthur:
Here is the follow up PR: https://github.com/reorg/pg_repack/pull/377

za-arthur:
This commit was merged together with the PR https://github.com/reorg/pg_repack/pull/377.

PR #162, opened 2017-12-19
Add support for --migration SQL statement

aks:
This PR adds a `--migration=SQLSTMT` (or `-m SQLSTMT`) option and implementation, which allows operations to be performed after the `CREATE TABLE` of the replacement table, but before the data `COPY` starts.

In other words, it's a perfect place to change the types of certain columns.  For example:

```
pg_repack -t big_table -m "ALTER COLUMN id TYPE BIGINT"
```
will change the ID column on the `big_table` to a `bigint`.

Currently, only a single STMT can be added.   Perhaps it might be useful to support multiple occurrences of `-m` and accumulate them all?

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -2266,7 +2266,7 @@ pgut_help(bool details)
 	printf("  -s, --tablespace=TBLSPC   move repacked tables to a new tablespace\n");
 	printf("  -S, --moveidx             move repacked indexes to TBLSPC too\n");
 	printf("  -o, --order-by=COLUMNS    order by columns instead of cluster keys\n");
-        printf("  -m, --migrate=SQLSTMT     possibly migrate one or more columns\n");
+        printf("  -m, --migrate=SQLSTMT     run SQLSTMT after CREATE TABLE\n");
 	printf("  -n, --no-order            do vacuum full instead of cluster\n");
 	printf("  -N, --dry-run             print what would have been repacked\n");
 	printf("  -j, --jobs=NUM            Use this many parallel jobs for each table\n");
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -194,6 +194,7 @@ typedef struct repack_table
 	const char	   *create_trigger;	/* CREATE TRIGGER repack_trigger */
 	const char	   *enable_trigger;	/* ALTER TABLE ENABLE ALWAYS TRIGGER repack_trigger */
 	const char	   *create_table;	/* CREATE TABLE table AS SELECT WITH NO DATA*/
+        const char         *migration_stmt;     /* optional ALTER TABLE .... stmt */
 	const char	   *copy_data;		/* INSERT INTO */
 	const char	   *alter_col_storage;	/* ALTER TABLE ALTER COLUMN SET STORAGE */
 	const char	   *drop_columns;	/* ALTER TABLE DROP COLUMNs */
@@ -228,6 +229,8 @@ static bool lock_exclusive(PGconn *conn, const char *relid, const char *lock_que
 static bool kill_ddl(PGconn *conn, Oid relid, bool terminate);
 static bool lock_access_share(PGconn *conn, Oid relid, const char *target_name);
 
+static char *temp_target_table_name(const repack_table *table);
+
 #define SQLSTATE_INVALID_SCHEMA_NAME	"3F000"
 #define SQLSTATE_UNDEFINED_FUNCTION		"42883"
 #define SQLSTATE_QUERY_CANCELED			"57014"
@@ -256,6 +259,8 @@ static bool				no_kill_backend = false; /* abandon when timed-out */
 static bool				no_superuser_check = false;
 static SimpleStringList	exclude_extension_list = {NULL, NULL}; /* don't repack tables of these extensions */
 
+static char                            *migration_stmt = NULL; /* optional statement */
+
 /* buffer should have at least 11 bytes */
 static char *
 utoa(unsigned int value, char *buffer)
@@ -276,6 +281,7 @@ static pgut_option options[] =
 	{ 's', 'o', "order-by", &orderby },
 	{ 's', 's', "tablespace", &tablespace },
 	{ 'b', 'S', "moveidx", &moveidx },
+        { 's', 'm', "migrate", &migration_stmt },
 	{ 'l', 'i', "index", &r_index },
 	{ 'b', 'x', "only-indexes", &only_indexes },
 	{ 'i', 'T', "wait-timeout", &wait_timeout },
@@ -773,6 +779,7 @@ repack_one_database(const char *orderby, char *errbuf, size_t errsize)
 	{
 		repack_table	table;
 		StringInfoData	copy_sql;
+                StringInfoData  migrate_sql;
 		const char *create_table_1;
 		const char *create_table_2;
 		const char *tablespace;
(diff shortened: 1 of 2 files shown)


dvarrazzo:
uhm... this cannot work? When you swap the file of the new table with the old table you have the metadata of the old table no?

As I understand how the system work, you find the old schema, but the data from the new schema... no?

jwg2s:
I would love to see this get added / merged. @aks @dvarrazzo how can I help? This would be a HUGE win for the postgres community to be able to perform zero-downtime migrations like this

dvarrazzo:
This doesn't work. Cannot work. Corrupts the server. Sorry to rain on your parade.

```
contrib_regression=# create table testaks (int serial primary key, date date, data text);
CREATE TABLE
contrib_regression=# insert into testaks values (default, current_date, 'whatever');
INSERT 0 1
contrib_regression=# select * from testaks;
 int |    date    |   data   
-----+------------+----------
   1 | 2018-11-16 | whatever
(1 row)


piro@makkuro:~/dev/pg_repack$ /usr/lib/postgresql/9.6/bin/pg_repack --table testaks --migrate="alter date type text" contrib_regression
INFO: repacking table "public.testaks"

piro@makkuro:~/dev/pg_repack$ psql contrib_regression
psql (9.6.11)
Type "help" for help.

contrib_regression=# \d testaks
                          Table "public.testaks"
 Column |  Type   |                       Modifiers                       
--------+---------+-------------------------------------------------------
 int    | integer | not null default nextval('testaks_int_seq'::regclass)
 date   | date    | 
 data   | text    | 
Indexes:
    "testaks_pkey" PRIMARY KEY, btree ("int")

contrib_regression=# select * from testaks;
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
Time: 7188.219 ms
?!> \q
piro@makkuro:~/dev/pg_repack$ 
```

Note that in the \d run after the repack the schema is reported as the one of the table before repacking: that's before the table is not "renamed": the data files of the repacked table are associated to the old schema. But the `--migrate` statement changes the table schema, creating data files inconsistent with the schema, and gracing ourselves with a server segfault.

As I said, this cannot work.

NikolayS:
The key question here is what causes Postgres to exit. It's worth to do a small research here, finding the root cause (setting log_min_messages to DEBUG, using strace and, finally, debugging with Valgrind https://wiki.postgresql.org/wiki/Valgrind).

The feature is a highly desired one and having it on plate would be really awesome.

NikolayS:
The key question here is what causes Postgres to exit. It's worth to do a small research here, finding the root cause (setting log_min_messages to DEBUG, using strace and, finally, debugging with Valgrind https://wiki.postgresql.org/wiki/Valgrind).

The feature is a highly desired one and having it on plate would be really awesome.

/cc @jwg2s @dvarrazzo

michail-nikolaev:
> As I understand how the system work, you find the old schema, but the data from the new schema... no?

Hello. But looks like it is possible to not alter pg_catalog to swap tables but just rename the tables with an exclusive lock (of course losing all FKs). What do you think?

shayonj:
ℹ️ Informational only (in case anyone else finds it useful - happy to retract)

I was recently looking into this as well, inspired by pg_repack and pg_squeeze. Here was my strategy for writing a new tool, but instead relying on the features from PG12 onwards for our schema change needs solves the issue for us. High level strategy - 

1. Create an audit table to record changes made to the parent table
2. Add a trigger on the parent table (for inserts, updates, deletes) to our audit table
3. (Acquire access exclusive lock on parent table) and create new shadow table with all rows from old table. 
4. Run ALTER/migration on the shadow table
5. Build indexes on the new table
6. Re-play all changes accumulated in the audit table ( will need to have control and knowledge of the type of schema change to make this operation safe)
    - Delete rows as they are replayed
7. Poll every 2 seconds on status and once the delta is ~20 rows, acquire a lock, within a transaction
   - swap table name (using `ALTER TABLE RENAME...`)
   - update references in other tables (FKs) by dropping and re-creating the FKs. 
8. Drop parent (now old) table

The major problem came from `by dropping and re-creating the FKs. `, since this would most likely negate the whole drill as doing that on another table could result in locks, etc making it not really zero downtime migrations/schema-changes. 

Anyways, I also found that PG 12 onwards we can rely on PG itself and take advantages of the native features to perform zero downtime migrations. For us - we saw issues (outages) mostly with the following kind of migrations and by using native alternatives from PG, we are able to solve for it without much of a heavy lift.

schmiddy:
> The major problem came from by dropping and re-creating the FKs. , since this would most likely negate the whole drill as doing that on another table could result in locks, etc making it not really zero downtime migrations/schema-changes.

On modern PG, I believe the dropping and re-creating of FKs could be done without needing long-lived Access Exclusive locks on the tables involved by dropping the constraints, re-adding them as `NOT VALID`, and then doing `ALTER TABLE ... VALIDATE CONSTRAINT`. It would be a bit slower than what pg_repack does, but could be done with plain SQL without needing the server-side extension. It'd be cool to script this up just to see how well it works (or maybe a tool like this already exists somewhere?)

shayonj:
Yeah, i thought of that as soon as I wrote this (from a separate thread). I think adding back FKs are doable. The `ALTER TABLE ... VALIDATE CONSTRAINT..` could be optional too. Depending on the setup, some ORMs can handle that as well. I plan on writing up something just to experiment. I have tried looking for alternative, but haven't come up with anything. Would def be curious. I know this getting a bit sidetracked from pg_repack's original purposes, but happy to jam more. I will link back, once I have something going (mostly next year). @shayonj on gmail / twitter (to avoid adding any noise to the thread) :)

phillbaker:
👋 hi all, I was interested in this topic as well and went down the path of forking pg_repack as discussed above and altering it to support schema changes. My proof of concept is here: https://github.com/phillbaker/pg_migrate, it's working for most cases, but hasn't been tested thoroughly and has some limitations (including renaming indexes).

In case other folks are curious and want to try something similar using pg_repack as a base, some interesting aspects that I found were:
* the create table statement for the copied table doesn't include defaults/nullable or a primary key
* indexes are renamed

On the other hand, I found that pg_repack's did provide a good base to build off of for the high level approach and insights into prior problems when doing this kind of work in the DB.

@shayonj I saw you started work on a CLI to do this in https://github.com/shayonj/pg-online-schema-change, maybe that's a better approach?

shayonj:
👋🏾 @phillbaker - very exciting to see more eyes on this problem. I skimmed through `pg_migrate` and it looks its building upon the [above strategy](https://github.com/reorg/pg_repack/pull/162#issuecomment-985612510) (adding / removing FKs, indexes, etc). Which I think makes sense. 

Yeah, I noticed pg_repack doesn't carry over the constraints. They are still preserved in the end since its updates the ref file nodes when doing the swap. 

Some of the limitations as `pg_migration` also exists in [`pg-online-schema-change`](https://github.com/shayonj/pg-online-schema-change#caveats--limitations). I think things like renamed indexes, constraints, sequences, are fine for now. I can think of a few ways to solve for it later. It does support dropped and renamed columns. It accordingly skips or copies data, when copying rows. 

I also want for `pg-online-schema-change` to ensure the change is reversible. That is, if after the swap, you want to go back to the original table, you can issue a new CLI command to do that, w/o any data loss. Mostly using triggers, there are come caveats but working on that as the last milestone. 

I plan on doing some stress testing with `pg-online-schema-change` and see how well it performs. I am still a bit skeptical on the last part where you need to drop and add constraints again. Since dropping FK constraints can end up holding locks for a while (if there is another transaction being performed) (?). Adding FK back is fine, since you can do `NOT VALID`.

shayonj:
Just FYI: The tool I talked about above is now live: https://github.com/shayonj/pg-online-schema-change

PR #166, opened 2018-01-17
Avoid excessive locking when table already being repacked by another instance

bashtanov:
When failing with "Another pg_repack command may be running on the table", pg_repack tries to perform some cleanup, that requires Access Exclusive lock on the table. It makes it wait for the first instance to complete, and prevents any usage of the table by third party until that, driving repack effectively locking like vacuum full. This PR is to fix it.

Steps to reproduce the problem:
1. Install postgres and repack, connect to postgres.
2. in psql create a table reasonably large to have some time to observe the issue while it's under repack:
```
create table rep_test(a serial, b text);
insert into rep_test(b) select repeat(' ', 100) from generate_series(1, 10000000);
alter table rep_test add primary key(a);
```
3. in bash run a few repacks (normally 2 is enough but 10 is bulletproof):
```
for i in `seq 1 10`; do pg_repack -t rep_test -E DEBUG 2>&1 | awk "{print \"REPACK$i\", \$0}" && echo $i & done
```
4. in psql:
```
select waiting, query from pg_stat_activity where state <> 'idle' and pid <> pg_backend_pid();
```

On master branch:
(3) All or most of the repack processes remain running.
(4) In addition to two working repack backends, there are a few waiting ones trying to `SELECT repack.repack_drop($1, $2)`

When fix applied:
(3) Only one repack remains running
(4) Only two its working backends present

### DIFF
--- a/lib/repack.c
+++ b/lib/repack.c
@@ -971,11 +971,21 @@ repack_drop(PG_FUNCTION_ARGS)
 	 * which in turn, is waiting for lock on log_%u table.
 	 *
 	 * Fixes deadlock mentioned in the Github issue #55.
+	 *
+	 * Skip the lock if we are not going to do anything.
+	 * Otherwise, if repack gets accidentally run twice for the same table
+	 * at the same time, the second repack, in order to perform
+	 * a pointless cleanup, has to wait until the first one completes.
+	 * This adds an ACCESS EXCLUSIVE lock request into the queue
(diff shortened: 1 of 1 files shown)


dvarrazzo:
@schmiddy ok to merge this? You analyzed #55 so I think you can make a more informed call. Thank you :)

schmiddy:
This looks good, verified it seems to help in the case of multiple `pg_repack` workers contending for the same table.

PR #172, opened 2018-03-09
Always qualify relation names.

MasahikoSawada:
Due to change at PostgreSQL 10.3, 9.6.8, 9.5.12, 9.4.17 and 9.3.22,
relation names passed by PostgreSQL function such as
pg_get_indexdef_string() are schema-qualified, which could be cause
of a parse error.

### DIFF
--- a/lib/pg_repack.sql.in
+++ b/lib/pg_repack.sql.in
@@ -16,6 +16,30 @@ CREATE FUNCTION repack.version_sql() RETURNS text AS
 $$SELECT 'pg_repack REPACK_VERSION'::text$$
 LANGUAGE SQL IMMUTABLE STRICT;
 
+CREATE FUNCTION repack.pg_version(version_str text DEFAULT NULL) RETURNS integer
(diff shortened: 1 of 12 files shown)


MasahikoSawada:
All Travis test passed.

In order to deal with differences between minor versions I added some new expected files and tricks. Any better idea is very welcome.

dvarrazzo:
@MasahikoSawada ouch, the boilerplate on top of every test file now has become really huge! I'm simplifying it a bit by moving by creating a `repack.pg_version()` function and using a different idiom, but I guess if it really gets in the way we can just drop that check and live with the problem we had before (i.e. identifying the files to change when tests changed)...

dvarrazzo:
I will merge this branch, then I guess we can work a bit on those tests files. We can isolate the variable parts (they are not many) into smaller files and drop the identity check.

MasahikoSawada:
Sorry for the late response. Agreed and thank you for merging!

PR #177, opened 2018-04-10
wrap tablespace name into quotes

kstep:
When a tablespace has some special characters (like hyphen "-"), the generated SQL produces syntax error. To avoid it, tablespace name should be wrapped into double quotes.

---
This change is [Reviewable](https://reviewable.io/reviews/reorg/pg_repack/177)

### DIFF
--- a/lib/repack.c
+++ b/lib/repack.c
@@ -770,7 +770,7 @@ repack_indexdef(PG_FUNCTION_ARGS)
 
 	/* specify the new tablespace or the original one if any */
 	if (tablespace || stmt.tablespace)
-		appendStringInfo(&str, " TABLESPACE %s",
+		appendStringInfo(&str, " TABLESPACE \"%s\"",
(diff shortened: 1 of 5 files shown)


dvarrazzo:
This patch is still broken if the namespace contains `"`. If you want to protect your dashes you should create a patch based on `quote_ident()`.

There is also no test showing the program didn't work before and works after your patch.

schmiddy:
And I think the current behavior is actually kind of intentional -- user-specified command-line parameters such as table name, index name, tablespace, etc. must be quoted appropriately by the user. This is admittedly odd behavior, but AFAIR it's exactly how command-line Postgres utilities such as `clusterdb`, `vacuumdb`, etc. work.

dvarrazzo:
Yes, I think identifier names are expected to be quoted and that this is uniformly the behaviour of postgres command line tool. Otherwise you couldn't tell if `foo.bar` is the table `bar` in the namespace `foo` or a table called `foo.bar` in a namespace on the path.

Closing

PR #188, opened 2018-09-18
Documentation : the wait timeout flag

magat:
I ran into a situation where the timeout was reached at the very beginning of the operations (because I use -D and there was an autovacuum running), and I was confused by the documentation.
The details section of the documentation states clearly that the ACCESS EXCLUSIVE locks are needed at the beginning and at the end of the repack, but I propose to make the `--wait-timeout`flag documentation a little clearer on this.

Thank you for this amazing tool btw !


xzilla (changes requested):
I think I would go a bit further... "needs to take one exclusive lock at the beginning and one exclusive lock at the end"

simi (approved):

PR #198, opened 2018-12-27
set idle_in_transaction_session_timeout to 0

NikolayS:
statement_timeout is being set to 0 in session preventing failures
when this parameter is non-zero. This change is to prevent failures due
to idle-in-transaction timeout, which look like this:

```
FATAL:  terminating connection due to idle-in-transaction timeout
ERROR: query failed: server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
DETAIL: query was: SAVEPOINT repack_sp1
```


NikolayS:
I don't know why tests don't go thru -- please review it in general. Looking forward to feedback.

dvarrazzo:
The tests fail with:

```

+cat regress/regression.diffs
--- /home/travis/build/reorg/pg_repack/regress/expected/nosuper.out	2018-12-28 00:43:59.323529663 +0000
+++ /home/travis/build/reorg/pg_repack/regress/results/nosuper.out	2018-12-28 00:44:29.928251196 +0000
@@ -14,4 +14,6 @@
 -- => ERROR
 \! pg_repack --dbname=contrib_regression --table=tbl_cluster --username=nosuper --no-superuser-check
 ERROR: pg_repack failed with error: ERROR:  permission denied for schema repack
+LINE 1: select repack.version(), repack.version_sql(), current_setti...
+               ^
 DROP ROLE IF EXISTS nosuper;
```

The test script does:

```
CREATE ROLE nosuper WITH LOGIN;
-- => OK
\! pg_repack --dbname=contrib_regression --table=tbl_cluster --no-superuser-check
```

It seems the error is not related to pg_repack, but to the database configuration for which a the user can't access the `repack` schema by default. I think after the `CREATE ROLE` we can add a `GRANT USAGE ON SCHEMA repack TO nosuper`, in the test and in the matching expected files.

NikolayS:
@dvarrazzo thank you. However the problem with tests is a bit more complicated – check out this https://travis-ci.org/reorg/pg_repack/jobs/477749374:
```
ERROR: pg_repack failed with error: You must be a superuser to use pg_repack
 -- => ERROR
 \! pg_repack --dbname=contrib_regression --table=tbl_cluster --username=nosuper --no-superuser-check
-ERROR: pg_repack failed with error: ERROR:  permission denied for schema repack
+ERROR: pg_repack failed with error: ERROR:  permission denied for relation tables
 DROP ROLE IF EXISTS nosuper;
+ERROR:  role "nosuper" cannot be dropped because some objects depend on it
+DETAIL:  privileges for schema repack
```

Any reasons to use a non-superuser role in tests?

NikolayS:
@dvarrazzo kindly pinging you about the questions above

dvarrazzo:
I'm sorry I don't have resources now to look into this.

andreasscherbaum:
@NikolayS Can you please rebase the PR? This will also trigger the GitHub Actions checks. Thanks.

PR #202, opened 2019-02-14
Don't suppress CREATE UNIQUE INDEX output

ChristophBerg:
Latest PostgreSQL minors (11.2 and friends) inhibit setting
client_min_messages greater than 'error'. The output from CREATE UNIQUE
INDEX has previously been ignored because it differs across versions,
but at least 9.4+ behave the same, so just proceed normally.

Context: https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=7b08b4a8a

### DIFF
--- a/regress/expected/repack-setup.out
+++ b/regress/expected/repack-setup.out
@@ -104,10 +104,9 @@ SELECT setseed(0); INSERT INTO tbl_with_mod_column_storage SELECT 1, array_to_st
  
 (1 row)
 
--- This will fail. Silence the message as it's different across PG versions.
-SET client_min_messages = fatal;
+\set VERBOSITY terse
 CREATE UNIQUE INDEX CONCURRENTLY idx_badindex_n ON tbl_badindex (n);
-SET client_min_messages = warning;
+ERROR:  could not create unique index "idx_badindex_n"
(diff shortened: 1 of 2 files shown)


Melkij:
Hi
currently pg_repack support even 9.1 postgresql: https://travis-ci.org/reorg/pg_repack
but i have no idea how make invalid index without error message now. So i think it is possible to drop community-unsupported versions now. In this case we need update Makefile (line 29), META.json, docs and possible something else.

ChristophBerg:
I haven't tested 9.1. There's chances that the included \set VERBOSITY terse makes the files compatible with versions older than 9.4.

Melkij:
hm, seems travis is happy for this patch even for 9.1. Maybe this comment was too old.

regress/expected/repack-setup.out
```diff
@@ -104,10 +104,9 @@ SELECT setseed(0); INSERT INTO tbl_with_mod_column_storage SELECT 1, array_to_st
  
 (1 row)
 
--- This will fail. Silence the message as it's different across PG versions.
```
MasahikoSawada: I think we should change only the second sentence, not remove whole comment.

ChristophBerg:
Fwiw, that was the patch I added to the Debian package to get it built. If more changes are necessary, please adjust, and release a new version so we can update the Debian package. Thanks!

MasahikoSawada:
Merged. Thank you!

PR #205, opened 2019-03-11
Escape table names for --only-indexes

panga:
Currently, the `--only-indexes` option doesn't escape table names given in the command line and causes the following error:

```
LOG: (query) LOCK TABLE table-name IN ACCESS EXCLUSIVE MODE
ERROR:  syntax error at or near "-"
LINE 1: LOCK TABLE table-name IN ACCESS ...
                                            ^
```

Example:
```
pg_repack -h SERVER -U USER -d DATABASE --echo --table='table-name' --only-indexes
```

Note: There's a workaround, you can add double quotes to the table name, however, this is something very easy to miss and cause the database become in an invalid state with a lot of indexes uncomplete rebuilt.

This PR change the SQL query used in `bin/pg_repack` and fix the issue.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -2132,7 +2132,7 @@ repack_all_indexes(char *errbuf, size_t errsize)
 	else if (table_list.head || parent_table_list.head)
 	{
 		appendStringInfoString(&sql,
-			"SELECT repack.oid2text(i.oid), idx.indexrelid, idx.indisvalid, idx.indrelid, $1::text, n.nspname"
+			"SELECT repack.oid2text(i.oid), idx.indexrelid, idx.indisvalid, idx.indrelid, repack.oid2text(idx.indrelid), n.nspname"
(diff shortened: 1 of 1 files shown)


schmiddy:
Thanks for the proposed fix - but this sounds like a duplicate of #177. Although the behavior is ugly it is intentional.

panga:
@schmiddy I understand. In this case, extra validation would be helpful to avoid issues and also some mention in the docs. Thank you

za-arthur (approved):
Thanks for the PR. It looks good to me.

> Thanks for the proposed fix - but this sounds like a duplicate of #177. Although the behavior is ugly it is intentional.

I think that #177 solves the issue in a different way by enforcing quotes. On the other hand this PR changes internal behavior of how table name is retrieved which is used later:
https://github.com/reorg/pg_repack/blob/999b854b5508ad29dc38e444f0898c41620eef9a/bin/pg_repack.c#L2191

On the other hand it is probably worth to add new tests.

PR #207, opened 2019-04-24
Support build for postgresql 12

Melkij:
Hello
I investigated into current PostgreSQL HEAD state (v12, feature freeze was few weeks ago) and found some changes incompatible with pg_repack. Here is my pull request with fix.

heap_open and heap_close was moved to [new header file](https://github.com/postgres/postgres/commit/4b21acf522d751ba5b6679df391d5121b6c4a35f) as part of new tableam API. I did not replace heap_open to relation_open - these names are still available as macro.

[New argument `is_index`](https://www.postgresql.org/message-id/E1gMxkM-0007jP-Li%40gemulon.postgresql.org) was added to `RenameRelationInternal`
I think it would be better define new RENAME_INDEX macro instead of adding new argument to RENAME_REL.

PostgreSQL 12 [no more supports](https://www.postgresql.org/message-id/E1gPG3j-0003cy-03%40gemulon.postgresql.org) tables with OIDS. And pg_class has no more relhasoids field.
I have no good idea how check field existence in query. Possible a better way would be define new C function with lookup in pg_class for older releases and noop for pg12. I did such change.

`make installcheck` works now. But I did't check tableam API, possible pg_repack works only with default heap tableam.

Thanks in advance!


realknorke:
I'm happe to help testing this patch when it is integrated into master and at least at "beta" level. ;)

lib/repack.c
```diff
@@ -112,8 +122,18 @@ must_be_superuser(const char *func)
...
+#else
+#define RENAME_REL(relid, newrelname) RenameRelationInternal(relid, newrelname, true, false);
+#endif
+/*
+ * is_index flag was added in 12.0, prefer separate macro for relation and index
+ */
+#if PG_VERSION_NUM < 120000
+#define RENAME_INDEX(relid, newrelname) RENAME_REL(relid, newrelname);
+#else
+#define RENAME_INDEX(relid, newrelname) RenameRelationInternal(relid, newrelname, true, false);
```
CyberDem0n: Maybe this is just nitpicking, but why the fourth argument is false if we use the `RenameRelationInternal()` to rename index?

Melkij: Ouch! Thank you, this is really my error.

Melkij:
> Since 12+ doesn't have OIDs anymore we could simply replace relhasoids with false

Great idea! I updated my PR in this way.
(maybe in april `create table ... with(oids=false)` was disallowed?..)

Also I add another tablespace test variant: seems since [this commit](https://github.com/postgres/postgres/commit/87259588d0ab0b8e742e30596afa7ae25caadb18#diff-640baf2937029728a8d51cccd554c2eeL1432) pg_get_indexdef_string does not returns pg_default tablespace and we have no test variant with qualified tablename and without `TABLESPACE pg_default`.

CyberDem0n:
> (maybe in april create table ... with(oids=false) was disallowed?..)

Nope. I think it will take about 5 years to remove it completely. 
Actually your old version of the patch with `get_relhasoids()` was just always generating `with (oids=false)`.

CyberDem0n:
And one more thing, which is not directly related to this rull request, but to the way how libraries a packaged on debian/ubuntu.
Right now it is not possible to build packages for postgres-11 and older if libpq-dev=12~beta3-1.pgdg18.04+1 is installed.
The build fails with following errors:
```
pg_repack.o: In function `lock_exclusive':
./bin/pg_repack.c:1818: undefined reference to `pg_snprintf'
./bin/pg_repack.c:1840: undefined reference to `pg_printf'
pg_repack.o: In function `utoa':
./bin/pg_repack.c:263: undefined reference to `pg_sprintf'
./bin/pg_repack.c:263: undefined reference to `pg_sprintf'
and so on
```

It happens because the https://github.com/postgres/postgres/commit/96bf88d52711ad3a0a4cc2d1d9cb0e2acab85e63#diff-95251b0ccd2b04c6807723d98aa8f6eeR137 is forcing to use postgres wrappers instead of a bunch of libc functions (snprintf and friends).

Now we have a weird situation, "version-independant" libpq-dev installs `/usr/include/postgresql/internal/port.h`, which is in fact a version dependant. pg_snprintf function and friends are located in the libpgport.a and libpgport_shlib.a, which are in turn belong to the `postgresql-server-dev-12` package...

cc: @df7cb

df7cb:
So far, all problems like that I've seen could be fixed by including the "server" headers first, before "internal" and the "client" headers.
Let me know if that's not the case here.

devrimgunduz:
ping ping ping.

dvarrazzo:
Can you update the documentation, mentioning support for PG 12 from 1.4.5, and make the tests on travis run on PG 12? https://travis-ci.org/reorg/pg_repack/builds/576432585

If so I'll merge and release.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 # Travis CI configuration file for psycopg2
 
-dist: trusty
```
dvarrazzo: If moving to bionic you find some old db unavailable, that's fine. Postgres atm only supports from 9.4 on and that's what we'd do.

Melkij: Both bionic and xenial fails during installation of some versions... I investigate...
Seems I can move to xenial with EOL db versions. But still have trouble with install of actual releases.

Melkij:
oh, success! travis is green now.

Melkij:
Thank you!

dvarrazzo:
Released: https://pgxn.org/dist/pg_repack/1.4.5/

Issue #210, opened 2019-07-16
Add support for PG12


realknorke:
Please add support for PG12.

I'm happy to help with testing the code afterwards. Thanks!

Melkij:
I already proposed patch in #207

realknorke:
Thanks @Melkij.

PR #213, opened 2019-09-02
Use new is_index argument of RENAME_REL() with versions above 12+

x4m:


x4m:
Uh, sorry, I see it duplicates #210

PR #231, opened 2020-03-30
Support for postgresql 13

Melkij:
Hello.
It is me again and my new patch for the upcoming PostgreSQL release v13.

- compatibility macroses heap_open and head_close [was removed](https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=f25968c49697db673f6cd2a07b3f7626779f1827) in postgresql.

This seems to be the only incompatible change so far.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Global excludes across all subdirectories
 *.o
 *.so
+*.bc
 regress/regression.diffs
 regress/regression.out
 regress/results/
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ dist: xenial
 sudo: required
 
 env:
+  - PGVER=13
+    PGTESTING=1
   - PGVER=12
   - PGVER=11
   - PGVER=10
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ sudo: required
 
 env:
   - PGVER=13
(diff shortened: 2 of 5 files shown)


dvarrazzo:
Hello, thank you for this.

Could you please add a note to the docs saying we will support PG 13 in next release and add postgres 13 on travis tests? https://travis-ci.org/github/reorg/pg_repack

Maybe it would be useful to test 13devel as a daily cronjob in addition to the normal grid (or only in the cronjob but not on push?)

Melkij:
Hm, package postgresql-13 should already exists in xenial-pgdg-testing. So let's try it. Previously, travis automatically ran tests on PR. Not sure why it's not works currently.

Postgresql 13 feature freeze is planned to April 7, 2020 AOE. So I didn't update the version number. We still have a small chance of new incompatible changes.

> Maybe it would be useful to test 13devel as a daily cronjob in addition to the normal grid
I think monthly cronjob would be a nice idea. I don’t think it makes sense to do more often

Melkij:
I have added commit with version bump along with the release notes. I guess we don't need to mention the changes of the URLs?
O, travis goes to check PR. Lets see...

Melkij:
Only 9.6 test is failed with unrelated error. For unknown reason freshly-installed postgresql was not started.

Melkij:
Thank you!

PR #253, opened 2020-11-01
Adds reference to docker image

hartmut-co-uk:
Hi, 
I'm maintaining a docker image for pg_repack - in the past I've had used it with a cron job for a scheduled 'cleanup' of tables with a high update frequency.
PR is as simple as referencing the docker image on the main README file - please advise if you'd rather have it someplace else?!?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,6 +5,7 @@ pg_repack -- Reorganize tables in PostgreSQL databases with minimal locks
 - Download: https://pgxn.org/dist/pg_repack/
 - Development: https://github.com/reorg/pg_repack
 - Bug Report: https://github.com/reorg/pg_repack/issues
+- Docker: https://hub.docker.com/repository/docker/hartmutcouk/pg-repack-docker
 
 |travis|
 


dvarrazzo:
I would prefer to avoid endorsing this docker image. If we had to release a docker image it should be released together with the release of an new pg_repack version, not left to an external entity. I also don't see too much of a use case for a docker image containing postgres + pg_repack: heavy postgres users, needing pg_repack, will have either postgres installed on a host, or if they use a docker image I would like to think they choose their version, they don't depend just on postgres:13.

As far as I can see I prefer to leave people to install their own database version and then pg_repack on top of it; this docker image has the postgres version decided to pg_repack, which is like the tail wagging the dog. Looking at your `install_pg_repack.sh` script it doesn't present any difficulty to do so.

hartmut-co-uk:
sure 👍 
One last note: the main use-case I've had back when I created the image was to easily 'invoke' pg_repack without having to manually install first. 
Also simplifies working with different versions (both of postgres/pg_repack).

PR #256, opened 2020-12-03
Travis-ci: added support for ppc64le

dthadi3:
Added power support for the travis.yml file with ppc64le. This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@
 dist: xenial
 sudo: required
 
+arch:
+  - AMD64
+  - ppc64le
+
 env:
   - PGVER=13
   - PGVER=12


dvarrazzo:
Hi @dthadi3 

Travis support for free-software projects has regressed to crap and is totally unreliable nowadays.

I have had too many problems with Travis and ARM to desire to bother: please see https://github.com/psycopg/psycopg2-wheels/issues/16 Even when it works, time to run jobs goes is in the several hours.

If you would like to send a merge request to run the pipeline on Github Actions it would be happily accepted.

Thank you very much!

dthadi3:
@dvarrazzo Thanks for your response, For more Info @gerrith3 can you please respond to his query.

gerrith3:
Hey @dvarrazzo is your project registered with travis-ci.ORG per chance?  Or have you migrated to travis-ci.COM - the production version of the Travis CI engine?  They are explicitly slowing down travis-ci.ORG because they announced it was going away a couple of years ago but people weren't migrating to the offiical site.  That would explain any slow down issues that you have seen.

As far as GHA - they don't support other architectures today - ppc64le, s390x, possibly not arm although that might have changed.  Anyway, we have a strong preference for Travis because it supports open source developers (https://blog.travis-ci.com/oss-announcement) and although there was a snafu when they switched to a credit based system because of the illegal and annoying use of travis by bitcoin miners and such, that should be straightened out now, and you can request additional credits from them via support.travis-ci.com if you don't have enough.  Things should be stabilizing now that the have completed the roll out of their next level of big changes.  I'd recommend ensuring you are tied to the .COM site (as opposed to .ORG) and see if things are better for you?

dvarrazzo:
@gerrith3 No, I think this project is still tied to the .org domain. Thank you for pointing it out.

I'll look into this migration and MR as soon as I have some capacity.

andreasscherbaum:
I have a GH Actions integration for the version we are using. It runs the regression tests, and also figures out if a PR is an upgrade, and then runs this upgrade as well. Obviously only on Ubuntu (one of the platforms supported by GH Actions).

Would this still be something which is useful for this repository? Maybe in addition to Travis?

andreasscherbaum:
We should just remove Travis support, right? GitHub checks are already used.

@dvarrazzo What do you think?

dvarrazzo:
Correct: Travis is no more in use.

andreasscherbaum:
Travis CI is no longer used.

andreasscherbaum:
PR to remove Travis CI: https://github.com/reorg/pg_repack/pull/344

PR #268, opened 2021-04-18
Fix the wrong URLs for the official home page

emres:
Fix the wrong URLs for the official home page and documentation page.


andreasscherbaum:
This was already changed. Thanks for the PR.

PR #275, opened 2021-06-01
WIP configurable upper bound on locking attempt timeout

ncleaton:
I would like to repack a busy table with minimal impact, and the hard-coded 1000ms statement timeout upper bound on exclusive lock attempts makes more disruption than I would like, by delaying latency-sensitive queries by up to 1 second. For this table, it would be better to have a very short timeout on exclusive lock attempts combined with many retries over a long period.

This patch is untested and lacking docs, but I think it has the right general shape. I've added a sleep so that reducing the lock timeout does not also make the lock retries more frequent.

Feedback please.


ncleaton:
I did some test runs, on a table with frequent small updates and lots of read traffic of mixed statement duration, with some read queries taking many seconds. Graphing the duration of the normally-quick statements shows the impact of the pg_repack. The scripts and raw data for these tests are at https://github.com/ncleaton/pg_repack_mr_275

With unpatched pg_repack I get the results graphed below. You can see the impact on normally very cheap queries of the multiple exclusive lock attempts at various stages of the pg_repack:

unpatched

With the first commit of this branch applied and `-m 10` in the options, it takes more tries to get the locks so the process takes longer, and the impact on normally fast queries lasts longer but is less severe. There's a short spike of slowness after the exclusive lock for the table swap is obtained, because the repack log has grown large during the long time it took to get the lock, so the final log replay while holding an exclusive lock takes a while:

first patch and -m 10

With the second patch applied there is less log to replay at the end, and I get:

both patches and -m 10

These runs are not very repeatable as there's a lot of luck in how long it takes to get each exclusive lock, but I haven't cherry picked them to make the patched versions look artificially good, honest.

andreasscherbaum:
@ncleaton Can you please rebase this patch? This will also trigger the checks.

bin/pg_repack.c
```diff
@@ -1948,6 +1961,10 @@ lock_exclusive(PGconn *conn, const char *relid, const char *lock_query, bool sta
 				pgut_rollback(conn);
 			else
 				pgut_command(conn, "ROLLBACK TO SAVEPOINT repack_sp1", 0, NULL);
+			if (timeout_msec < wait_msec)
+				usleep(1000 * (wait_msec - timeout_msec));
+			if (apply_log_table)
+				apply_log(conn, apply_log_table, 0);
```
za-arthur, line 1967: It is a good idea to apply log during attempts of acquiring lock to minimize duration of ACCESS EXCLUSIVE lock.

Although I'm trying to think of a way to avoid adding new argument `apply_log_table` in addition to `relid` and `lock_query`. 
Having just only these arguments would be nice:
```
static bool
lock_exclusive(PGconn *conn, const repack_table *table, bool start_xact, bool apply_log)
```

In most cases instance of `repack_table` already has relation OID and lock query. Except this case:
https://github.com/reorg/pg_repack/blob/5ae252fdd20e6866994fa30865b8c5b5e3d3b9b4/bin/pg_repack.c#L2181

andreasscherbaum: @ncleaton Any idea about the suggestions from @za-arthur?

ncleaton: Hmm, to make that work repack_table_indexes would need to construct a repack_table struct, which seems confusing  when the table itself is not being repacked. I don't think it's justified to reduce the arg count by one.

ncleaton:
I rebased and replied, I don't think this is "waiting for author" any more

ncleaton:
oh it needed rebasing again for another conflict :man_facepalming:

Issue #276, opened 2021-06-04
PostgreSQL 14 support


devrimgunduz:
Hi,

CREATE EXTENSION fails against v14:

[postgres] # CREATE EXTENSION pg_repack ;
ERROR:  function array_append(anyarray, anyelement) does not exist

Can you please take a look? v14 is already in beta state.

Regards, Devrim

Melkij:
Sure, I am working on a pull request. I am late this year, but my patch still need more polishing before publishing. This is not the only thing that has been changed during v14 developing cycle.

devrimgunduz:
Thanks @Melkij ! I can confirm that the package builds with the patch. Will wait for the new release.

Regards, Devrim

devrimgunduz:
@Melkij any updates? 14.0 is due soon.

Melkij:
@devrimgunduz , I am not a committer...

devrimgunduz-edb:
@schmiddy @dvarrazzo ping

and @Melkij sorry!

dvarrazzo:
@devrimgunduz-edb Is EDB interested in giving any support to the project?

dvarrazzo:
@Melkij as you have been actively helping the project I would have no problem in giving you commit power, if you would like.

Melkij:
@dvarrazzo , thanks. I use pg_repack at work and plan to continue making it work with newer postgresql releases.

dvarrazzo:
Thank you @Melkij. Added as maintainer. Thank you very much for your work so far! :+1:

Melkij:
@dvarrazzo , thank you, now I have the commit power. I am going to merge #277 in a few days. Just to be sure: do I need to do anything besides tagging and creating [a release here](https://github.com/reorg/pg_repack/releases/new) to publish a new release?

dvarrazzo:
@Melkij there is [a document](https://github.com/reorg/pg_repack/blob/master/doc/release.rst) with the release instructions. Once packages are ready I can upload them on PGXN.

Melkij:
@dvarrazzo, right, I've definitely seen it before. I merged #277 pull request. According to the documentation, the new package should be uploaded into pgxn before tagging, so I stopped on this point.

bilva-sanaba:
Thank you for all the contributions @Melkij. Just checking in @dvarrazzo, are there any other changes intended to go in before the 1.4.7 tag. If not, is there any timeline for this tag?

devrimgunduz:
[...]

Issue #63, opened 2016-01-14
pg_repack compiling fail on Ubuntu 14.04 with apt.postgresql.org packages


Slach:
pg_repack not compatible with libpq5 9.5 ???

steps to reproduce
<pre>
wget -q -O- https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
echo "deb http://apt.postgresql.org/pub/repos/apt/ trusty-pgdg main" > /etc/apt/source.list.d/pgsql.list
apt-get update
apt-get install libedit-dev libpq-dev postgresql-9.4 postgresql-server-dev-9.4 
pip install --upgrade pgxnclient
pgxn install pg_repack
</pre>

error output 
<pre>
root@local-coffee-master:~# pgxn install pg_repack
INFO: best version: pg_repack 1.3.2
INFO: saving /tmp/tmpcBVeFA/pg_repack-1.3.2.zip
INFO: unpacking: /tmp/tmpcBVeFA/pg_repack-1.3.2.zip
INFO: building extension
make[1]: Entering directory `/tmp/tmpcBVeFA/pg_repack-1.3.2/bin'
gcc -Wall -Wmissing-prototypes -Wpointer-arith -Wdeclaration-after-statement -Wendif-labels -Wmissing-format-attribute -Wformat-security -fno-strict-aliasing -fwrapv -fexcess-precision=standard -g -g
-O2 -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -I/usr/include/mit-krb5 -fPIC -pie -DLINUX_OOM_SCORE_ADJ=0 -fno-omit-frame-pointer -I/usr/include/postgresql -DREPACK_V
ERSION=1.3.2 -I. -I./ -I/usr/include/postgresql/9.4/server -I/usr/include/postgresql/internal -D_FORTIFY_SOURCE=2 -D_GNU_SOURCE -I/usr/include/libxml2  -I/usr/include/tcl8.6  -c -o pg_repack.o pg_repa
ck.c
In file included from pgut/pgut.h:22:0,
                 from pgut/pgut-fe.h:13,
                 from pg_repack.c:25:
/usr/include/postgresql/internal/pqexpbuffer.h:149:13: warning: no previous prototype for 'printfPQExpBuffer' [-Wmissing-prototypes]
 extern void printfPQExpBuffer(PQExpBuffer str, const char *fmt,...) pg_attribute_printf(2, 3);
             ^
[...]

PR #277, opened 2021-06-07
Support for postgresql 14

Melkij:
Hello
I think I am ready to publish a new patchset to support the postgresql 14 release. I found several changes that we need to address:

- SQL function `array_append` was changed. Most noticeable part, already mentioned in #276 . Instead of fixing the `repack.array_accum` custom aggregate function, I want to drop it and use native PostgreSQL string_agg and array_agg functions instead. They are both old enough.
- API of `simple_prompt` function was changed again
- I replaced few SearchSysCacheCopy calls to encouraged macro. This fixes segfault on pg14 for me, but still not sure why exactly.
- postgresql 14 start producing `CONTEXT:  unnamed portal parameter $2 = '...'` in case of `--table relation_not_exists`. Looks weird. I added explicit check for tables existence
- also I temporary enabled `PGVER=9.6` test on travis. Just for check.

Thanks in advance!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,15 @@ dist: xenial
 sudo: required
 
 env:
+  - PGVER=14
+    PGTESTING=1
   - PGVER=13
   - PGVER=12
   - PGVER=11
   - PGVER=10
   # Disabled because packages broken at least on xenial
   # https://www.postgresql.org/message-id/CA%2Bmi_8a1oEnCzkt0CvqysgY4MQ6jEefjmS%3Dq_K-AvOx%3DF7m2%2BQ%40mail.gmail.com
-  # - PGVER=9.6
+  - PGVER=9.6
   - PGVER=9.5
   - PGVER=9.4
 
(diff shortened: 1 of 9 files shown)


Melkij:
Great! It works! Sorry about the noise, otherwise I can't check Travis.
So basically I reverted `regress/expected/repack-check.out` to make tests pass on older postgresql branches. We still have `ERROR: pg_repack failed with error: ERROR:  relation "dummy_table" does not exist` message.
It looks like the PGDG apt repo now requires to specify `(os)-pgdg main 14` instead of` (os)-pgdg 14` as we did earlier. (In hope I did not break #63).

PS: also I noticed `Please consider migrating to travis-ci.com.` message on travis. Take a look, please

dvarrazzo:
> PS: also I noticed Please consider migrating to travis-ci.com. message on travis. Take a look, please

It would be better to move to Github Actions instead. I've done that for almost all my other repositories.

Issue #285, opened 2021-08-25
pg_repack Leaks Passwords to Logs on Connection Error


danielmerken:
On connection error, the following is printed:

> ERROR: could not connect to database with "dbname=******** password=******** ": could not connect to server: Connection refused

I've redacted the dbname and password, but they are printed in plaintext to the log. I noticed this in version 1.4.5.

dvarrazzo:
You know that you can use the PGPASSWORD env var, right?

mikkorantalainen:
As it's expected that connection / info string contains credentials, it should not be ever logged. Log the extracted info (e.g. database name should not be a secret) that doesn't contain credentials. The environment variable can be used *as a workaround* but it's not a correct fix for this issue.

danielmerken:
We initially noticed this issue when running PG repack using the PGPASSWORD env var, so I don't believe it avoids this issue.

andreasscherbaum:
#286 fixes this issue.

PR #286, opened 2021-08-26
Remove connection info from error log

danielmerken:
Removing connection info from error log, which can include sensitive info such as DB password. 

Addresses this issue: https://github.com/reorg/pg_repack/issues/285


danielmerken:
@schmiddy what would it take to get this released? We'd like to have this issue resolved as soon as possible so we don't risk leaking creds.

PR #291, opened 2021-11-16
remove unnecessary plus operator

lincuiping:
I found an unnecessary plus operator in bin/pgput/pgut.c
I remove it.


Melkij:
Thanks!

Issue #281, opened 2021-07-07
Why doesn't pg_repack support `ALTER EXTENSION ... UPDATE` syntax? 


Valeria1235:
Hi,

The documentation says that pg_repack doesn't support `alter extension ... update`, which means that I would need to drop & create the extension if I were to upgrade it.
Are there any good reasons why this feature is lacking? Would you accept a patch adding all (or more realistically, a subset covering last 4-5 major versions) of the missing migrations?

dvarrazzo:
The extension is stateless. Is there a valid use case for upgradability and the extra burden of creating migrations?

Valeria1235:
This causes inconsistency when updating the extension in systems like SaltStack. You need to make a special case for updating pg_repack.

dvarrazzo:
That's way less of the effort it takes to maintain an upgradeable extension. Writing a stateful salt command for the task is trivial.

devrimgunduz:
@dvarrazzo upgrades are breaking systems badly, until someone drops and recreates the extension:

pg_repack failed with error: extension 'pg_repack 1.4.7' required, found extension 'pg_repack 1.4.6'

Why don'y you just run a CREATE OR REPLACE FUNCTION ... between 1.4.6 and 1.4.7 as an upgrade script?

I'm not too familar with repack code, so I'd better ask you.

dvarrazzo:
Hi @devrimgunduz 

pg_repack is not upgradable and it's difficult to make it so, because it generates dynamically triggers, functions, maybe other structures: the upgrade script should upgrade not only pg_repack code but also the generated artifacts.

Not only it is difficult, but *it makes no sense to do so* because pretty much pg_repack has no state, except during repack, which is definitely a moment in which you don't want to upgrade the extension...

So, I don't consider:

> pg_repack failed with error: extension 'pg_repack 1.4.7' required, found extension 'pg_repack 1.4.6'

as "breaking the system badly". All that it has to be done is to drop the extension and re-create it. We could write a more reassuring error message but nothing is broken here.

PR #294, opened 2021-12-06
Reassure the user that it's ok to drop the extension

dvarrazzo:
See #281


Melkij:
No objection from me. Feel free to merge.

PR #295, opened 2021-12-07
Use the correct article for abbreviations

sunwei-ch:
We've accumulated quite a mix of instances of "an SQL" and "a SQL" in the
documents.  It would be good to be a bit more consistent with these.

The most recent version of the SQL standard I looked at seems to prefer
"an SQL".  That seems like a good lead to follow, so here we change all
instances of "a SQL" to become "an SQL".  Most instances correctly use
"an SQL" already, so it also makes sense to use the dominant variation in
order to minimise churn.

### DIFF
--- a/bin/pgut/pgut-fe.c
+++ b/bin/pgut/pgut-fe.c
@@ -573,7 +573,7 @@ parse_pair(const char buffer[], char key[], char value[])
 }
 
 /*
- * execute - Execute a SQL and return the result.
+ * execute - Execute an SQL and return the result.
  */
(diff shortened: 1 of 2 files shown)


schmiddy:
> Most instances correctly use "an SQL" already, so it also makes sense to use the dominant variation in
order to minimise churn.

I actually don't see any existing mentions of "an SQL" in the code. Seems to be standardized as-is on "a SQL" so I'd just as soon leave this alone.

```
$ git grep -i "an SQL"
$
```

PR #298, opened 2022-01-06
Take ACCESS EXCLUSIVE LOCK carefully

za-arthur:
Try to get the lock in loop with timeouts using `lock_exclusive()`. This allows to not stuck lock queue for readers if pg_repack cannot take the lock fast enough.

We had incidents when cleanup functions, which run `repack.repack_drop()`, couldn't get ACCESS EXCLUSIVE lock in a reasonable time. It was because there was long running write connections, running around 5-10 minutes. `repack.repack_drop()` also stucks lock queue block any new readers until all old writes finish.


za-arthur:
@andreasscherbaum I rebased the PR.

Issue #299, opened 2022-01-07
Entering a blank password causes infinite loop


egashira-yusuke:
pg_repack hangs when entering an empty or blank string at the password prompt of pg_repack.
It does not return until interrupt with Ctrl+C.
Also, during this, pg_repack is continuously trying to connect to the database.
This causes outputting a large amount of connection log if log_connections=on was specified in the database.
I found this problem in PostgreSQL 14.1 and pg_repack 1.4.7.

[Reproduce procedure]
1. Create a password authenticated user on the database.
2. Run pg_repack specifing 1.'s user. (Do not specify -w option)
3. When the password prompt appears, press Enter without entering anything or with entering only blank.

According to my research, this problem seems to start from #184.
In pgut_connect()'s connection loop, if PQconnectionNeedsPassword(conn) is true, 
the password prompt is appeared by prompt_for_password() and it requires input of password. 
prompt_for_password() remembers the password.
In pgut_connect(), the password obtained from prompt_for_password() is appended to the connection string as "password=%s" format.
Then, the connection string will end with "password=" because the entered password was empty string.
On the other hand, PostgreSQL Server seems to request the password again if the connection string ends with "password=".
Therefore, PQconnectionNeedsPassword(conn) is still true during reconnect loop of pgut_connect(), and it causes infinite loop.

MichaelDBA:
Validated the hanging condition...

andreasscherbaum:
Fixed in #354

PR #303, opened 2022-03-10
Fix infinite loop when an empty or blank password is given

egashira-yusuke:
This is fix for #299.

The connection string used by pg_repack was missing password quotes.
It also adds retry processing so that if an empty password is entered, the password prompt does not remember it.

### DIFF
--- a/bin/pgut/pgut-fe.c
+++ b/bin/pgut/pgut-fe.c
@@ -63,7 +63,7 @@ setup_workers(int num_workers)
  		if (username && username[0])
  			appendStringInfo(&buf, "user=%s ", username);
  		if (password && password[0])
- 			appendStringInfo(&buf, "password=%s ", password);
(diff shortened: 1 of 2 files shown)


mvdobrinin:
This should help with #315.

za-arthur (commented):
Thanks for the PR. What about instead of another infinite loop like here:
```
do {
    buf = simple_prompt("Password: ", BUFSIZE, false);
} while (strlen(buf) == 0);
```

just fail with error `no password supplied` after `prompt_for_password()` returned empty string?
This will be similar behavior to psql's.

andreasscherbaum:
@egashira-yusuke Since this PR is fixing an issue with empty passwords: can you add a test which shows that empty passwords are no longer an issue?

andreasscherbaum:
Fixed in #354 instead.

PR #309, opened 2022-05-19
support for postgresql 15

Melkij:
Postgresql 15 beta1 is here. Luckily, we only have one minor change: postgresql commit 98e93a1fc93e9b54eb477d870ec744e9e1669f34 removed include of pwd.h and breaks build. I decided to replace your own implementation of get_username and use the generic get_user_name_or_exit from the postgresql code. Exists since postgresql 9.4, so no change in supported versions.

### DIFF
--- a/bin/pgut/pgut-fe.c
+++ b/bin/pgut/pgut-fe.c
@@ -9,6 +9,7 @@
 
 #define FRONTEND
 #include "pgut-fe.h"
+#include "common/username.h"
 
 #ifdef HAVE_GETOPT_H
 #include <getopt.h>
@@ -16,7 +17,7 @@
 #include <getopt_long.h>
 #endif
 
-char	   *dbname = NULL;
+const char *dbname = NULL;
 char	   *host = NULL;
 char	   *port = NULL;
 char	   *username = NULL;
@@ -33,8 +34,6 @@ worker_conns workers   = {
 
 
 static bool parse_pair(const char buffer[], char key[], char value[]);
-static char *get_username(void);
-
 
 /*
  * Set up worker conns which will be used for concurrent index rebuilds.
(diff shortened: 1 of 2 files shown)


dvarrazzo:
Does tests still run somewhere?

Melkij:
I doubt it.

capri1989:
Are there any plans to merge this PR to add PostgreSQL 15 support? The GA release of PG 15 is only a week ahead, so just wondering what the state here is as this topic seems stale.

capri1989:
Maybe that was just me being too excited, I just looked at the GA releases over the last few years, which usually happened end of Q3. Thanks for getting back to me that quickly, really appreciated.

df7cb:
> Even the RC1 date [has not yet been announced](https://wiki.postgresql.org/wiki/PostgreSQL_15_Open_Items#Important_Dates), where did you find announce of GA in a week ahead?
> 
> I probably merge this patch after checking on RC1 manually.

RC1 is due tomorrow, and 15.0 on October 13th.

Melkij:
So, I merged this patch (and the neighboring bugfix) and tested it on versions from 9.4 to 15RC1 (+ HEAD)
I also committed the preparation of release 1.4.8. The pgxn magic remains.

take-five:
Is it possible to backport it to older versions of pg_repack? We use AWS RDS and they don't update pg_repack for older versions of PostgreSQL: https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-extensions.html

For example, Postgres 12 only supports 1.4.5, but now it's no longer possible (or very difficult) to compile pg_repack 1.4.5 against current Postgres on Alpine.

df7cb:
pg_repack currently supports PG 9.5+ (at least that's what the Debian package tells me, it's been a while since 9.5 was built), so just building pg_repack for PG12 should be trivial for Amazon. You just need to get them to do that.

PR #316, opened 2022-07-26
Create pg_repack-14.spec

cvas71:


Melkij:
Please complement the description - what is this?

cvas71:
Spec file for creating rpm(Red Hat Package Manager) package for PostgreSQL 14.

dvarrazzo:
That's the packager's responsibility, not upstream's.

The previous spec file is a relic of the past which can be dropped from the project.

PR #318, opened 2022-09-28
Add github workflow

df7cb:
This runs the regression tests as a GitHub action.

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -1,0 +1,45 @@
+name: make installcheck
+on: [push, pull_request]
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        pg:
+          - 15
+          - 14
+          - 13
+          - 12
+          - 11
+          - 10
+
+    name: PostgreSQL ${{ matrix.pg }}
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


df7cb:
How it looks like: https://github.com/credativ/pg_repack/actions/runs/3144576352

dvarrazzo:
@df7cb tests were cancelled at the first failure. Could you please add `fail-fast: false` to the grid? See [here](https://github.com/psycopg/psycopg/blob/013a5a44cde065f82ca63e7181db1d3e6f925a73/.github/workflows/tests.yml#L23-L26) for instance.

I would also remove the unsupported versions from the grid and starting from v10.

df7cb:
Ah, good idea with disabling fail-fast, added in the last push. I had 9.5/9.6 in there because they work, but also removed them now.

Example run: https://github.com/credativ/pg_repack/actions/runs/3158761985

Melkij:
Looks good to me.

PS: repack should pass the tests even on 9.4, I just checked it.

dvarrazzo:
Merged, thank you!

Melkij:
pg15 needs libzstd-dev, I pushed an update

dvarrazzo:
Thank you, @Melkij, I was just about to ask :)

PR #324, opened 2022-10-31
Bump checkout github action version

fabriziomello:
All github actions that run on Node12 are deprecated so bumped `checkout` github action version to run on Node16.

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


dvarrazzo:
Merged, thank you!

PR #326, opened 2022-11-17
Add a where clause

ankitml:
hardcoded for now. 
add a input flag.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -944,7 +944,13 @@ repack_one_database(const char *orderby, char *errbuf, size_t errsize)
 		/* Craft Copy SQL */
 		initStringInfo(&copy_sql);
 		appendStringInfoString(&copy_sql, table.copy_data);
-		if (!orderby)
+
+        /* soft delete recognition */
+        appendStringInfoString(&copy_sql, " WHERE ");
+        appendStringInfoString(&copy_sql, "deleted_at IS NULL");
+//        table.create_table = sql.data;
+
+        if (!orderby)
 
 		{
 			if (ckey != NULL)


ankitml:
@fabriziomello curious to know what are your thoughts on this

andreasscherbaum:
> It could cause data loss if used incorrectly

How can this be prevented?

ankitml:
Thanks, removed the workspace file

ankitml:
> > It could cause data loss if used incorrectly
> 
> How can this be prevented?

The arguments to where clause as specified by the user is intended to delete data. Users would need to verify if data being deleted is what they want to delete. Dry run of the pg-repack shows precisely which data is going to be remain. It is recommended to think through the outputs of dry run before running real command.

ankitml:
We have been running this repack at instacart on few tables weekly to clean up zombie tuples for last 4 months. Without any manual intervention, on a cron.

andreasscherbaum:
@ankitml Can you add unit tests to ensure this is working as intended?
Please also add tests which have ... let's say unusual where clauses. Like something which is broken, which clearly should not be in a where clause, something with spaces or special characters. Anything which can break the query.

ankitml:
@andreasscherbaum Added basic tests but not failure ones. The current error if a bad input is given in the INSERT query. ie the bulk insert that happens at the start of repack. This is not ideal. Ill rework repack.c so it catches error beforehand and raises.

ankitml:
@andreasscherbaum Added few things
1. Constraint for using this flag only for single table repacks (just to protect user from unforced errors), added a test to catch it
2. Validation checks for where clause by running an explain select 1 from table where (WHERE_CLAUSE). If an invalid where clause is added, it fails here. 
3. If validation fails, parse postgres error and convey the error message to user by pointing that where clause is invalid. 
4. Added tests for few invalid cases. 
5. Docs clearly mentioning where clause should contain condition that determines what rows to keep. (ie the reverse condition decides what gets deleted). 

Let me know how it looks, anything else that needs to be done for it to be code complete.

bin/pg_repack.c
```diff
@@ -2445,8 +2471,97 @@ pgut_help(bool details)
...
+    
+    res = PQexec(conn, sql.data);
+    
+    if (PQresultStatus(res) != PGRES_TUPLES_OK)
+	/* Clean up the error message. Remove duplicate ERROR: prefix as well as remove postgres LINE: error */
+    {
+        if (errbuf)
+        {
+            const char *pg_error = PQerrorMessage(conn);
+            char cleaned_error[1024] = {0}; // Buffer for cleaned error message
```
ankitml, line 2514: Now, since the underlying error message comes from postgres. this could lead to overflows. I am not sure how to handle it, perhaps make it 10x to almost reduce chances of an overflow.

za-arthur:
Hey @ankitml,

There are few concerns about the current where clause support:
- AFAIK the where clause is not applied to the trigger function and doesn't handle a log table. That implementation might be tricky since the trigger function `repack.repack_trigger()` is a security definer function and it would be good to avoid SQL injections.
- A table can have foreign keys references. Copying less records which were on the original table can break integrity.

Thoughts?

ankitml:
> AFAIK the where clause is not applied to the trigger function and doesn't handle a log table. That implementation might be tricky since the trigger function repack.repack_trigger() is a security definer function and it would be good to avoid SQL injections.
Yes, this would mean the incoming changes to the table while the table is repacking are not pruned. I have mentioned this in the `what this PR doesnt do`. Making this delete operation a best effort prune than a strict prune. Honestly, I think this is okay as long as it is communicated to the user. Most of the times developers already have a soft delete functionality in their application and this repack's job will be to clean up those soft deleted tuples entirely. The cleanup does not need to be strict. 

>A table can have foreign keys references. Copying less records which were on the original table can break integrity.
Thats a good point. This will be not ideal outcome. Let me think what can be done. Do you have suggestions? Our internal use cases this is done by the soft-delete functionality in the application so all rows that are pruned by repack have this already handled but this cannot be guaranteed out in the wild.

andreasscherbaum:
> Most of the times developers already have a soft delete functionality in their application

This is not the only way to create bloat in a table, by far. You have a soft delete in your application, and this is causing bloat. Without evidence I'd say it's not even close to "most of the time" that developers use this approach.

ankitml:
I dont see an easy way out for foreign keys. Making this feature potential for misuse/accidents. 😭

andreasscherbaum:
I share these concerns.

PR #327, opened 2022-12-20
Handling system tables errors

chanukya571:
added a function check_systemtables() which will check whether its a system table or not.

if its a system table it will throw a meaningful error to use VACUUM FULL.


bin/pg_repack.c
```diff
@@ -487,6 +529,12 @@ preliminary_checks(char *errbuf, size_t errsize){
 		goto cleanup;
 	}
 
+	if (check_systemtables()) {
```
za-arthur, line 532: Maybe it would make sense to make this check inside `repack_one_table` and `repack_table_indexes` only? That way this check won't interrupt pg_repack completely, and it will skip only system tables.

za-arthur (changes requested):

PR #328, opened 2022-12-29
Handling tables with storage plain

chanukya571:
code for handling when tables having all columns  storage plans are PLAIN.


chanukya571:
Checks failed

PR #336, opened 2023-03-07
Flag for stopping repack when invalid index is found

EinKrebs:
Currently, when we find an incorrect index, we only show warning. With --error-on-invalid-index flag, there'll be an error and repack will be stopped. It's done to prevent corruption of indexes. Index may be corrupted if repack is interrupted, then index is updated, and then repack is done completely.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -1325,8 +1325,8 @@ repack_one_table(repack_table *table, const char *orderby)
 		const char *indexdef;
 		indexdef = getstr(indexres, j, 0);
 		if (error_on_invalid_index) {
-			elog(ERROR, "Invalid index: %s", indexdef);
-			return;
+			elog(WARNING, "Invalid index: %s", indexdef);
+			goto cleanup;
 		} else {
 			elog(WARNING, "skipping invalid index: %s", indexdef);
 		}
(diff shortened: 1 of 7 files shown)


andreasscherbaum:
@EinKrebs Can you please rebase the patch, and fix the conflict? Thanks!

bin/pg_repack.c
```diff
@@ -1321,7 +1324,12 @@ repack_one_table(repack_table *table, const char *orderby)
 	{
 		const char *indexdef;
 		indexdef = getstr(indexres, j, 0);
-		elog(WARNING, "skipping invalid index: %s", indexdef);
+		if (error_on_invalid_index) {
+			elog(ERROR, "Invalid index: %s", indexdef);
+			return;
```
za-arthur, line 1329: Instead of calling `elog(ERROR)` and `return` it is better to call `elog(WARNING)` and `goto cleanup`.
Here is an example:
https://github.com/reorg/pg_repack/blob/a1489a1e6c15be8913ce37674a0adc74184c013f/bin/pg_repack.c#L1421

za-arthur (changes requested):

andreasscherbaum:
LGTM

Issue #313, opened 2022-06-20
Failure to repack a table with all columns with PLAIN storage


vmarkovtsev:
I've got this table:
```sql
CREATE TABLE github.node_ref (
    name text,
    prefix text,
    acc_id bigint NOT NULL,
    associated_pull_requests_ecnt bigint,
    graph_id bigint NOT NULL,
    repository_id bigint,
    target_id bigint
);
ALTER TABLE ONLY github.node_ref ALTER COLUMN name SET STORAGE PLAIN;
ALTER TABLE ONLY github.node_ref ALTER COLUMN prefix SET STORAGE PLAIN;
ALTER TABLE ONLY github.node_ref
    ADD CONSTRAINT node_ref_pkey PRIMARY KEY (acc_id, graph_id);
```
Therefore, the two `text` columns should not get TOAST-ed.
`pg_repack` fails with the following log:
<details>
<pre>
06:41:03  [repack github.node_ref]
06:41:03  repack jobs: 1
06:41:03  validating connection
06:41:05  LOG: (query) SET search_path TO pg_catalog, pg_temp, public
06:41:05  LOG: (query) SET search_path TO pg_catalog, pg_temp, public
06:41:05  LOG: (query) select repack.version(), repack.version_sql()
06:41:05  LOG: (query) SET statement_timeout = 0
06:41:05  LOG: (query) SET search_path = pg_catalog, pg_temp, public
06:41:05  LOG: (query) SET client_min_messages = warning
06:41:05  LOG: (query) SELECT r FROM (VALUES ($1)) AS given_t(r) WHERE NOT EXISTS(  SELECT FROM repack.tables WHERE relid=to_regclass(given_t.r) )
06:41:05  LOG: 	(param:0) = github.node_ref
06:41:05  LOG: (query) SELECT t.*, coalesce(v.tablespace, t.tablespace_orig) as tablespace_dest FROM repack.tables t,  (VALUES (quote_ident($1::text))) as v (tablespace) WHERE (relid = $2::regclass) ORDER BY t.relname, t.schemaname
06:41:05  LOG: 	(param:0) = (null)
06:41:05  LOG: 	(param:1) = github.node_ref
06:41:05  INFO: repacking table "github.node_ref"
06:41:05  LOG: (query) SELECT pg_try_advisory_lock($1, CAST(-2147483648 + $2::bigint AS integer))
06:41:05  LOG: 	(param:0) = 16185446
06:41:05  LOG: 	(param:1) = 17808
[...]

PR #337, opened 2023-03-15
Add support for tables, which have been previously rewritten with VACUUM FULL and have Storage for all columns set to PLAIN

kotik-adjust:
As explained in https://github.com/reorg/pg_repack/issues/313, pg_repack is unable to repack tables, which have been previously rewritten with VACUUM FULL and have Storage for all columns set to PLAIN

Example error: `ERROR: query failed: ERROR:  repack_swap : unexpected toast relations (T1=0, I1=0, T2=68003958, I2=68003960`

This PR tries to add support for this case.


lib/pgut/pgut-spi.c
```diff
@@ -84,6 +84,8 @@ execute_with_args(int expected, const char *src, int nargs, Oid argtypes[], Datu
 	int		i;
 	char	c_nulls[FUNC_MAX_ARGS];
 
+    	memset(c_nulls, 0, sizeof(c_nulls));
```
kotik-adjust, line 87: W/o this change tests are failing with `error: ‘c_nulls’ may be used uninitialized [-Werror=maybe-uninitialized]`

Issue #338, opened 2023-03-15
Support for all old EOL'd PG versions and solving all racing conditions?


MichaelDBA:
Currently, the pg_repack documentation supports PostgreSQL versions:
PostgreSQL 9.4, 9.5, 9.6, 10, 11, 12, 13, 14, 15

I recommend stopping support for old EOL'd versions ( > 1 yr old ) so that supported versions at the present time would be:
PostgreSQL 10 (EOL), 11, 12, 13, 14, 15.

Even though it might still work for these 2yr and older versions, let's state that it is not guaranteed to work and no support will be provided.  Supporting the latest EOL version seems reasonable enough.

There's too much work here already with complicated changes needed to accomplish complicated things like racing conditions, which I think we should shy away from anyway.  Why do you think VACUUM FULL has no enhancement path to add the CONCURRENTLY option?

Personally, I think we should not keep reaching for the sky in trying to resolve all racing conditions.  Perhaps add some coding logic to detect heavy update activity with respect to a table being vacuumed and give up after some specified time.

dvarrazzo:
No problem in stopping support for Postgres < 10 or 11.

As per the race conditions problem, they are difficult reproduce and to test. As far as I understand, it is a problem that the core team doesn't feel to solve either, which is the reason why this project has never really been considered to be included in core.

Melkij:
Well, I don't remove old versions if pg_repack passes tests without any changes. I don't have a plan to support certain versions. As long as they don't interfere, I leave it as is.

> Why do you think VACUUM FULL has no enhancement path to add the CONCURRENTLY option?

I know: because no one is working on it. It's a long process to bring the discussion to a consensus, implement the patch and bring it to merge to the mainline. A lot of patches are stopped due to lack of attention during review.

andreasscherbaum:
> Well, I don't remove old versions if pg_repack passes tests without any changes.

I agree, as long as old versions are working, pg_repack can - and should - be tested on them.
Only if problems arise with old versions then they can be made EOL, and the documentation should be updated to reflect this.

MichaelDBA:
Nice feedback, stuff to think about...

PR #345, opened 2023-04-03
support for postgresql 16

Melkij:
Hello

So far there are no incompatible changes, 1.4.8 works with postgresql 16dev without any changes.

However, I propose to prepare a new release 1.4.9 as usual, I will mention postgresql 16 support in the change notes to eliminate any ambiguity.

Question for discussion: it is necessary to change the list of supported old versions? ( #338 )

PS: at the same time I want to check if pg16 is already available for our tests

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -7,6 +7,7 @@ jobs:
       fail-fast: false
       matrix:
         pg:
+          - 16
           - 15
           - 14
           - 13


Melkij:
>  update-alternatives: error: alternative path /usr/share/postgresql/16/man/man1/postmaster.1.gz doesn't exist
> dpkg: error processing package postgresql-16 (--configure):
>  installed postgresql-16 package post-installation script subprocess returned error exit status 2

The package is available, but its installation fails. I'll trigger the check few weeks later.

andreasscherbaum:
This is currently failing with an unrelated error in the package:

```

File: META.json [hunk 1]
```diff
@@ -2,7 +2,7 @@
    "name": "pg_repack",
    "abstract": "PostgreSQL module for data reorganization",
    "description": "Reorganize tables in PostgreSQL databases with minimal locks",
-   "version": "1.4.8",
+   "version": "1.4.9",
```
andreasscherbaum, line 5: I would like to separate the update of support for PG 16 and the cut of a new release. These two do not necessarily have to go into a single PR.

Melkij:
I removed unrelated changes and ran tests locally. For postgresql 16 we didn't need any additional changes. This time, the github tests also passed without surprises.

franciscoreinolds:
Greetings!

Any news in regards to this? 👀

Melkij:
@franciscoreinolds , pg_repack 1.4.8 works fine with postgresql 16. No code changes were required. Regarding the mention of pg16 support in the documentation and tests... I lost the track.

@andreasscherbaum , I remember that a new release was planned.

franciscoreinolds:
Indeed, I meant more if there had been a discussion to release a new version with the content of this PR just to explicitly support PG 16 👌🏽

andreasscherbaum:
Yes, there is a bugfix we want to get in, and then release a version which mentions 16.

.github/workflows/regression.yml
```diff
@@ -7,6 +7,7 @@ jobs:
       fail-fast: false
       matrix:
         pg:
+          - 16
```
dvarrazzo, line 10: The file `doc/pg_repack.rst` must be updated too to declare the next version pg16-compatible.

dvarrazzo (changes requested):

za-arthur:
This was done by https://github.com/reorg/pg_repack/pull/364 and https://github.com/reorg/pg_repack/pull/372

PR #261, opened 2020-12-24
new option "switch-threshold"

gsmol:
Hello!
Recently, when running pg_repack on really busy table, we were unable to catchup due to MIN_TUPLES_BEFORE_SWITCH been less than the size of average insert, populating the table.
So I would like to propose the patch, which should remedy such situations by parameterizing the switch threshold value instead of using macro. New options is "-r | --switch-threshold"

Thank you for a great tool.

PR #347, opened 2023-04-13
Rebase: new option "switch-threshold"

za-arthur:
This is rebased patch from https://github.com/reorg/pg_repack/pull/261.

Differences:
- There is no unnecessary short option name `-r`
- Default threshold value is `100`
- New test for `--switch-threshold`

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -2386,5 +2386,5 @@ pgut_help(bool details)
 	printf("  -Z, --no-analyze          don't analyze at end\n");
 	printf("  -k, --no-superuser-check  skip superuser checks in client\n");
 	printf("  -C, --exclude-extension   don't repack tables which belong to specific extension\n");
-	printf("  -t, --switch-threshold    switch tables when that many tuples are left to catchup\n");
+	printf("      --switch-threshold    switch tables when that many tuples are left to catchup\n");
 }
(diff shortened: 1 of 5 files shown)


bin/pg_repack.c
```diff
@@ -287,6 +288,7 @@ static pgut_option options[] =
 	{ 'b', 'k', "no-superuser-check", &no_superuser_check },
 	{ 'l', 'C', "exclude-extension", &exclude_extension_list },
 	{ 'b', 'F', "error-on-invalid-index", &error_on_invalid_index },
+	{ 'i', 1, "switch-threshold", &switch_threshold },
```
Melkij, line 291: I'm confused by the meaning of such a constant here. What does the value 1 mean here? It is not a `required_argument from` getopt_long

za-arthur: Basically it is a unique identifier of the option. It should be unique and other options shouldn't use it. Other option is to use `indexptr` of `getopt_long`, but pg_repack doesn't use it.
Options represented by alphabetic symbols (short option name) also shouldn't be same when you convert it to integer.
https://www.gnu.org/software/libc/manual/html_node/Getopt-Long-Options.html

pg_repack uses its own wrapper `struct pgut_option` around `getopt_long`. And second value is `sname`:
https://github.com/reorg/pg_repack/blob/944dbf7a49adfed7bd2ad467ae9431e8067442a4/bin/pgut/pgut-fe.h#L42

pg_repack copies `struct pgut_option` into `struct option` here:
https://github.com/reorg/pg_repack/blob/944dbf7a49adfed7bd2ad467ae9431e8067442a4/bin/pgut/pgut-fe.c#L656

andreasscherbaum (approved):

andreasscherbaum:
@za-arthur Please rebase this PR.

PR #349, opened 2023-04-14
fix #200 lock_timeout instead of statement_timeout

### LINKED ISSUE #200: Canceled "RESET statement_timeout" due to statement_timeout=100ms
On a busy server we've got this multiple times:

```bash
Jan 10 16:51:33 ERROR: query failed: ERROR:  canceling statement due to statement timeout
Jan 10 16:51:33 DETAIL: query was: RESET statement_timeout
```

I assume that `SET LOCAL statement_timeout = 100ms` exists only for the subsequent `LOCK TABLE ..` command, doesn't it? It is not expected that `RESET` might be so long – but it happened.

If so, isn't it better to use `SET LOCAL lock_timeout ...` instead? `RESET` will not be limited by it, while for `LOCK TABLE ..` the same effect as with setting `statement_timeout` will be achieved.

tometzky:
On a busy database, if statement_timeout=100, then "RESET statement_timeout" can take more than 100ms and fail. So use lock_timeout instead, available since PostgreSQL 9.3.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -231,7 +231,7 @@ static bool lock_access_share(PGconn *conn, Oid relid, const char *target_name);
 
 #define SQLSTATE_INVALID_SCHEMA_NAME	"3F000"
 #define SQLSTATE_UNDEFINED_FUNCTION		"42883"
-#define SQLSTATE_QUERY_CANCELED			"57014"
+#define SQLSTATE_LOCK_NOT_AVAILABLE		"55P03"
 
 static bool sqlstate_equals(PGresult *res, const char *state)
 {
@@ -1798,7 +1798,7 @@ lock_access_share(PGconn *conn, Oid relid, const char *target_name)
 
 		/* wait for a while to lock the table. */
 		wait_msec = Min(1000, i * 100);
-		printfStringInfo(&sql, "SET LOCAL statement_timeout = %d", wait_msec);
+		printfStringInfo(&sql, "SET LOCAL lock_timeout = %d", wait_msec);
 		pgut_command(conn, sql.data, 0, NULL);
 
 		printfStringInfo(&sql, "LOCK TABLE %s IN ACCESS SHARE MODE", target_name);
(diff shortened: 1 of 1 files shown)


andreasscherbaum:
@tometzky Does it solve the underlying problem? Also when other LOCKs are held on the table?
And what is a good way to test the new functionality?

tometzky:
@andreasscherbaum 

> Does it solve the underlying problem?

It did for us. We have not encountered timeout errors on "reset" command on a busy database after applying the patch.

> Also when other LOCKs are held on the table?

The statement_timeout has a wider scope than lock_timeout, and includes lock_timeout. In the [setting documentation](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-LOCK-TIMEOUT) there's "if statement_timeout is nonzero, it is rather pointless to set lock_timeout to the same or larger value, since the statement timeout would always trigger first". Also this value only is relevant for the time needed to acquire a lock - it does not affect locks already acquired in the transaction, so no other locks can be relevant.

> And what is a good way to test the new functionality?

If you mean that how to reproduce the problem and then test if the patch will fix it, then I don't think I'm able to do this. It would require either forcing a test database to perform "reset" for longer than 100ms, or somehow "mock" responses from the database.

And if you mean to test the currently implemented retry manually, if it works correctly with lock_timeout at least as well as with statement_timeout, then I'd try to run:
```
create table test_pg_repack as select (random()*10^12)::bigint as id, counter
    from generate_series(1,(10*10^6)::int) as counter;
alter table test_pg_repack add primary key (id, counter);

do $$
begin
    loop
        perform 1 from test_pg_repack;
        perform pg_sleep(1);
        commit;
    end loop;
end;
$$;
```


This would likely (about 90% of the time) force pg_repack to retry locking the table, as 100ms it tries first, will likely not be enough. Here's an example log with the patch: https://upload.syncron.com/RpObEJcJyyx66KhBWSpA/pg_repack.log

But, as this is timing related, and not 100% reliable, I think it might be quite hard to create a reliable unit test from this. At least not without a very significant effort to "mock" database responses.

andreasscherbaum:
We decided to merge this. Since the last pg_repack release is a while ago, this deserves a longer RC period.

andreasscherbaum:
The `lock_timeout` appeared first in [PostgreSQL 9.3](https://www.postgresql.org/docs/9.3/runtime-config-client.html).

scorpevans (approved):

PR #350, opened 2023-04-17
Support larger password lengths

maratik123:
Limit 100 seems too low for long passwords

### DIFF
--- a/bin/pgut/pgut.c
+++ b/bin/pgut/pgut.c
@@ -442,7 +442,7 @@ prompt_for_password(void)
 	static char *passwdbuf;
 	static bool have_passwd = false;
 
-#define BUFSIZE 100
+#define BUFSIZE 1024
 
 #if PG_VERSION_NUM < 100000
 	if (have_passwd) {


za-arthur:
Thanks for the pull request.

Looks like Postgres 14 removed password length limit at all from its tools:
https://github.com/postgres/postgres/commit/67a472d71c98c3d2fa322a1b4013080b20720b98

Would you like improve the PR and remove password length limit as well if `pg_repack` is being built with Postgres 14+?

dvarrazzo:
Why should the limit be removed only if building with 14+ Do we use postgres code to manage password reading?

If that's what we do, and it's easy to handle longer passwords using such code, no problem in removing the limit only using pg14+. Otherwise, if we need to write our own code, it shouldn't need to be conditional.

Also,

> Limit 100 seems too low for long passwords

not to me. If there has to be a limit somewhere, this limit seems more than adequate. If we want to drop the limit that's ok, but considering a password chosen in 64^99 = 2^594 combinations "insecure" has my sincere lols.

dvarrazzo:
(you can always add a punctuation symbol).

Melkij:
> Do we use postgres code to manage password reading?

yep, all simple_prompt function variants are from postgresql frontend utils: https://github.com/reorg/pg_repack/blob/master/bin/pgut/pgut.c#L439

za-arthur:
> If that's what we do, and it's easy to handle longer passwords using such code, no problem in removing the limit only using pg14+. Otherwise, if we need to write our own code, it shouldn't need to be conditional.

That is correct. On the other hand it wouldn't be hard to write the code, the function `simple_prompt` can be copied as-is.

https://github.com/postgres/postgres/blob/REL_15_2/src/common/sprompt.c#L53

maratik123:
> > Do we use postgres code to manage password reading?
> 
> yep, all simple_prompt function variants are from postgresql frontend utils: https://github.com/reorg/pg_repack/blob/master/bin/pgut/pgut.c#L439

also need to support PGPASSWORD env variable to pass passwords in automated pg_repack execution

maratik123:
Close in favor to #357

za-arthur:
Thanks @maratik123 for the PR. It was merged together with https://github.com/reorg/pg_repack/pull/357.

PR #354, opened 2023-05-31
Check if a password already was prompted

za-arthur:
This fixes infinite loop in case of an empty password.
The issue #299.

There is another PR #303 which fixes the issue. But it adds another infinite loop and it isn't consistent with the behavior when a user inputs wrong password.

In case of a wrong password pg_repack throws an error like:
```
ERROR: could not connect to database: connection to server on socket "/tmp/.s.PGSQL.5432" failed: FATAL:  password authentication failed for user "repack_user"
```
and it doesn't go to another loop and doesn't ask for another password.

In case of an empty password with this PR pg_repack will throw an error like:
```
ERROR: could not connect to database: connection to server on socket "/tmp/.s.PGSQL.5432" failed: fe_sendauth: no password supplied
```

### DIFF
--- a/bin/pgut/pgut.c
+++ b/bin/pgut/pgut.c
@@ -542,10 +542,9 @@ pgut_connect(const char *info, YesNo prompt, int elevel)
 			return conn;
 		}
 
-		if (conn && PQconnectionNeedsPassword(conn) && prompt != NO)
+		if (conn && PQconnectionNeedsPassword(conn) && !passwd && prompt != NO)
 		{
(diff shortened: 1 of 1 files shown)


andreasscherbaum:
Does this need a test to make sure it works in the future?

za-arthur:
It would be a bit tricky to create a test because I'm not sure how to provide a password to pg_repack in tests. It expects a user input.
AFAIK using pipe like the following doesn't work:
```
echo "test" | pg_repack ...
```

PR #357, opened 2023-06-07
Support larger password lengths v2

za-arthur:
This PR enhances https://github.com/reorg/pg_repack/pull/350. In addition to increasing of buffer size it removes restrictions on password length for Postgres 14+.

As it was discussed starting from [this comment](https://github.com/reorg/pg_repack/pull/350#issuecomment-1516559391) it is possible to remove the restriction completely by copying `simple_prompt` function from the Postgres 14+ source code into pg_repack code base. But this looks more complicate than removing the restriction only for Postgres 14+, because it will be necessary to support that code in the future.

It is possible to remove the restriction completely in the future if it will be decided that it is better.

### DIFF
--- a/bin/pgut/pgut.c
+++ b/bin/pgut/pgut.c
@@ -442,7 +442,7 @@ prompt_for_password(void)
 	static char *passwdbuf;
 	static bool have_passwd = false;
 
-#define BUFSIZE 100
+#define BUFSIZE 1024
 
 #if PG_VERSION_NUM < 100000
 	if (have_passwd) {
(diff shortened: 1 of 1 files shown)


bin/pgut/pgut.c
```diff
@@ -467,13 +471,15 @@ prompt_for_password(void)
 	}
 #else
 	if (have_passwd) {
-		buf = pgut_malloc(BUFSIZE);
-		memcpy(buf, passwdbuf, sizeof(char)*BUFSIZE);
+		buf = pgut_malloc(passwd_size);
```
maratik123, line 474: `passwd_size` is equal to 0?

za-arthur: Nope, `passwd_size` is static variable and the value is assigned here:
https://github.com/reorg/pg_repack/blob/905fd9f1931a3b78619ccb4eb25e57516ce7eaa8/bin/pgut/pgut.c#L478

PR #360, opened 2023-06-21
Support for exclude table and exclude parent table option

chanukyasds:
Idea is to support pg_repack with below options

1. exclude-table (-X) : which skips a table from repacking
2. exclude-parent-table (-Y): which skips a table and its children from repacking

once its merged I will be writing the regression tests.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -244,6 +244,7 @@ static bool				noorder = false;
 static SimpleStringList	parent_table_list = {NULL, NULL};
 static SimpleStringList	table_list = {NULL, NULL};
 static SimpleStringList exclude_table_list = {NULL, NULL};
+static SimpleStringList exclude_parent_table_list = {NULL, NULL};
 static SimpleStringList	schema_list = {NULL, NULL};
 static char				*orderby = NULL;
 static char				*tablespace = NULL;
@@ -289,6 +290,7 @@ static pgut_option options[] =
 	{ 'b', 'k', "no-superuser-check", &no_superuser_check },
 	{ 'l', 'C', "exclude-extension", &exclude_extension_list },
 	{ 'l', 'X', "exclude-table", &exclude_table_list },
+	{ 'l', 'Y', "exclude-parent-table", &exclude_parent_table_list },
 	{ 'b', 2, "error-on-invalid-index", &error_on_invalid_index },
 	{ 'i', 1, "switch-threshold", &switch_threshold },
 	{ 0 },
@@ -362,7 +364,7 @@ main(int argc, char *argv[])
 	}
 	else
 	{
-		if (schema_list.head && (table_list.head || parent_table_list.head || exclude_table_list.head))
+		if (schema_list.head && (table_list.head || parent_table_list.head || exclude_table_list.head || exclude_parent_table_list.head ))
 			ereport(ERROR,
 				(errcode(EINVAL),
 				 errmsg("cannot repack/exclude specific table(s) in schema, use schema.table notation instead")));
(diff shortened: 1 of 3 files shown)


andreasscherbaum:
I'm voting for not using short options for this feature, just the long ones.

@za-arthur @Melkij What do you think?

Melkij:
Personally, I don't use pg_repack without `-t` at all, only with a specific table of a specific database. But there are no objections from me. These options do not overcomplicate the logic and it looks like they might be useful for someone. (maybe, for consistency, we also need --exclude-schema?)

Yes, I agree that it is better to add only long options. It would be better to have the -T option for --exclude-table (the pg_dump analogy would be very handy), but that option was already taken.

chanukyasds:
I agree -T is not available.

I think it's a better idea that we all can discuss and re-arrange options with best analogy.

Even I thought of below options for better understanding.
 
 -t   --table
 -T  --parent-table
 
 -x --exclude-table
 -X --exclude-parent-table
 
 --exclude-schema is a good option. I will work on that and create a pull soon.

za-arthur:
I agree that we probably want to add new short option names only if it is necessary and it is a good idea to add long names only for this PR.
We have limited number of short names and it is better to use them wise.

bin/pg_repack.c
```diff
@@ -600,13 +628,27 @@ is_requested_relation_exists(char *errbuf, size_t errsize){
 		if (iparam < num_relations)
 			appendStringInfoChar(&sql, ',');
 	}
+	for (cell = exclude_table_list.head; cell; cell = cell->next)
```
za-arthur, line 631: This logic is probably unnecessary since `exclude-table` conflicts with `table` and `parent-table`:
https://github.com/reorg/pg_repack/pull/360/files#diff-a9d4d5b8968d339c50f089a6b531b1f36eedd27cdc008bdb26b70fb9fd0eee2fR377

bin/pg_repack.c
```diff
@@ -600,13 +628,27 @@ is_requested_relation_exists(char *errbuf, size_t errsize){
...
+			appendStringInfoChar(&sql, ',');
+	}
 	for (cell = parent_table_list.head; cell; cell = cell->next)
 	{
 		appendStringInfo(&sql, "($%d, 'p')", iparam + 1);
 		params[iparam++] = cell->val;
 		if (iparam < num_relations)
 			appendStringInfoChar(&sql, ',');
 	}
+	for (cell = exclude_parent_table_list.head; cell; cell = cell->next)
```
za-arthur, line 645: Same as above `exclude-parent-table` conflicts with `table` and `parent-table`:
https://github.com/reorg/pg_repack/pull/360/files#diff-a9d4d5b8968d339c50f089a6b531b1f36eedd27cdc008bdb26b70fb9fd0eee2fR382

bin/pg_repack.c
```diff
@@ -357,11 +364,26 @@ main(int argc, char *argv[])
 	}
 	else
 	{
-		if (schema_list.head && (table_list.head || parent_table_list.head))
+		if (schema_list.head && (table_list.head || parent_table_list.head ))
```
za-arthur, line 367: Can you remove this white space?

za-arthur (commented):
Generally the PR looks good to me as well. I have couple comments below.
In the addition to those comments I'd suggest you to add new tests to test new options.

bin/pg_repack.c
chanukyasds: Same as above.

This logic helps pg_repack to find the exclude-parent-table exists in the database or not.

For example consider below case.

`pg_repack -p 5433 -d postgres  -Y public.ptable2`

In above command user can skip a parent table from a database

bin/pg_repack.c
chanukyasds: Cleared the white space.

chanukyasds:
I have removed short option names and provided long option names only.

@andreasscherbaum  @Melkij  @za-arthur

PR #372, opened 2023-11-22
Add PostgrerSQL16 into supported versions

za-arthur:
Add PostgrerSQL16 into supported versions into `docs/pg_repack.rst` and `docs/pg_repack_jp.rst`.
Tests were added by https://github.com/reorg/pg_repack/pull/364.

We probably could delete oldest PostgreSQL versions which are not listed in https://github.com/reorg/pg_repack/blob/6263dd5e7fcb2890c68b4237591628c4ad08259d/.github/workflows/regression.yml#L10-L16

### DIFF
--- a/doc/pg_repack.rst
+++ b/doc/pg_repack.rst
@@ -40,7 +40,7 @@ Requirements
 ------------
 
 PostgreSQL versions
-    PostgreSQL 9.4, 9.5, 9.6, 10, 11, 12, 13, 14, 15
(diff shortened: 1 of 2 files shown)


andreasscherbaum (approved):
lgtm

nodakai:
https://reorg.github.io/pg_repack/ doesn't mention PG 16 yet (and I had to use GitHub search to confirm that PG 16 is indeed supported.)

za-arthur:
I created a PR https://github.com/reorg/reorg.github.com/pull/3 to reflect this change.

PR #373, opened 2023-11-23
Fix regression tests

dbonne:
* Removed regression tests for postgres 10 and 11
* Fixed regression tests


.github/workflows/regression.yml
```diff
@@ -12,14 +12,11 @@ jobs:
           - 14
           - 13
           - 12
-          - 11
-          - 10
```
andreasscherbaum, line 16: Is there a reason to remove the tests? Yes, PostgreSQL version 10 and 11 is no longer supported, but it will be nice to know if and when `pg_repack` fails on these versions.

andreasscherbaum:
What exactly is fixed in the regression tests?

CyberDem0n:
It is not really a fix, but IMO an improvement, because it removes the need to create a tablespace outside of regression tests by utilizing `allow_in_place_tablespaces` GUC, which was back-ported down to 10.22.

But, I agree, there is no reason not to run tests on 10 and 11.

PR #374, opened 2023-11-28
Issue #338: Remove old EOL PostgreSQL versions

za-arthur:
Currently we test only Postgresql from 10 to 16:
https://github.com/reorg/pg_repack/blob/6263dd5e7fcb2890c68b4237591628c4ad08259d/.github/workflows/regression.yml#L10-L16

And we cannot guarantee that older PostgreSQL versions work correctly.

Issue: https://github.com/reorg/pg_repack/issues/338


andreasscherbaum:
Objection: the Unit Tests are not failing, which implies that `pg_repack` is working on the older versions.
The tests on older versions should be kept as an indication as to when `pg_repack` is failing on older PostgreSQL versions.

If we add a feature which breaks compatibility with older versions, there's no point in fixing that. But that's the time we can update the documentation and clearly state that we know it is no longer working with version xyz. If tests for older versions are removed now, there is no information as to if the tool works or not.

Melkij:
Reasonable. We probably also need to remove obsolete regress/travis_prepare.sh and regress/travis_test.sh

andreasscherbaum:
I agree on the Travis part.

Melkij:
I remember that there was a problem installing such old versions of postgresql for autotesting. I'm actually a little surprised that the pg10 test still works on github workflows. I think it will stop working in the spring (after the next ubuntu-latest is released).

za-arthur:
I'll add old postgres versions into the workflow to see which versions fail.

za-arthur:
By ede9bc3e8e76b8e801ecd11612b6e79f0b290f6b I fixed tests for Postgres 9.6, 9.5. Postgres couldn't get `create_table` column type and showed warning:

```
  CREATE EXTENSION pg_repack;
+ WARNING:  column "create_table" has type "unknown"
+ DETAIL:  Proceeding with relation creation anyway.
```

And therefore `init-extension` tests output didn't match with expected output.

Now it seems Postgres 9.4 fails because of compilation errors. I think it isn't worth to try to fix them and I will delete Postgres 9.4 from tests.

andreasscherbaum:
> Postgres 9.4

Please also update the docs at the same time stating that it is known that 9.4 no longer works.

That's the whole point of keeping the old tests around.

za-arthur:
> Please also update the docs at the same time stating that it is known that 9.4 no longer works.

Done, without translation to Japanese.

andreasscherbaum (approved):
Looks good now.

PR #375, opened 2023-12-14
Fix wrong OID format type in repack_trigger

Melkij:
Table OID should be unsigned four-byte integer, therefore formatting with %d is wrong. %u should be used as in other places.

It's surprising that we haven't received such error messages before. But such a query is definitely wrong: 
```
INSERT INTO repack.log_-907750756(pk, row) ...
```

I have no idea how to add a test with large OID.


za-arthur (approved):
Thank you that makes sense to me. LGTM.

Melkij:
Thank you for review

Issue #282, opened 2021-08-02
must have a primary key or not-null unique keys when INCLUDE columns


pguser4ever:
pg_repack version 1.4.6

Running pg_repack on table (in this case a partition) generate an error with pk and INCLUDE column

pg_repack -h dbserver1 
--no-kill-backend  --table=dummy.parttable2_2021_07_13 -j 4 --order-by=sdrid,sdrtimestamp

NOTICE: Setting up workers.conns
WARNING: relation "dummy.parttable2_2021_07_13" must have a primary key or not-null unique keys

pk: 
PRIMARY KEY, btree (sdrid, sdrtimestamp) INCLUDE (applicationserverid)   

reorg tables without include columns works fine 

If we add NOT NULL to INCLUDE column i primary key it works 
My assumption was that INCLUDE columns is not a part of the primary key and that pg_repack should just ignore 
the INCLUDE column/columns in this case and only use pk columns for sorting

sgrinko:
We really need support for this opportunity!
What is the status for now?

za-arthur:
Should be fixed by https://github.com/reorg/pg_repack/pull/376

PR #376, opened 2023-12-14
Issue 282: Fix check of NOT NULL by repack.primary_keys

za-arthur:
PostgreSQL 11 introduced covering indexes. pg_repack couldn't use indexes which include additional columns because all columns are listed in `pg_index.indkey`. To overcome this it is necessary to consider `pg_index.indnkeyatts` value.

In `lib/pg_repack.sql.in` I had to define two versions of `repack.primary_keys` since `pg_index.indnkeyatts` was added only in PostgreSQL 11. Otherwise it would be necessary to drop support of PostgreSQL 10 and below.

Issue: https://github.com/reorg/pg_repack/issues/282

### DIFF
--- a/META.json
+++ b/META.json
@@ -2,7 +2,7 @@
    "name": "pg_repack",
    "abstract": "PostgreSQL module for data reorganization",
    "description": "Reorganize tables in PostgreSQL databases with minimal locks",
-   "version": "1.5.0",
+   "version": "1.5.1",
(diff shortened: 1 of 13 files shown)


andreasscherbaum (approved):
Looks good

andreasscherbaum:
I like that you keep older versions supported.

PR #377, opened 2024-01-05
Fix swap of relfrozenxid, relfrozenxid and relallvisible

za-arthur:
This PR is follow up of https://github.com/reorg/pg_repack/pull/157.

- I fixed the typo: https://github.com/reorg/pg_repack/pull/157#issuecomment-1519195513
- I removed `PG_VERSION_NUM >= 90300` check since we don't support such old versions anyway
- I added check `relform1->relkind != RELKIND_INDEX` since `relfrozenxid` and `relminmxid` is non-zero only for tables

I think we still need to check the current state of `swap_relation_files` (https://github.com/postgres/postgres/blob/ae69c4fcf1337af399a788ab8d96af540bd1cd8e/src/backend/commands/cluster.c?plain=1#L1055). It has a few differences with the code of `repack.c`. We don't need to copy all the code, since not all features are supported by pg_repack. For example we don't repack system tables, mapped tables. And maybe we don't need to support `swap_toast_by_content`?

I'll try to come up with a PR to try to sync the code.

### DIFF
--- a/lib/repack.c
+++ b/lib/repack.c
@@ -1189,7 +1189,6 @@ swap_heap_or_index_files(Oid r1, Oid r2)
 	Form_pg_class relform1,
 				relform2;
 	Oid			swaptemp;
-	TransactionId swaptempxid;
 	CatalogIndexState indstate;
 
 	/* We need writable copies of both pg_class tuples. */
@@ -1231,13 +1230,16 @@ swap_heap_or_index_files(Oid r1, Oid r2)
 	 */
 	if (relform1->relkind != RELKIND_INDEX)
 	{
-		swaptempxid = relform1->relfrozenxid;
+		TransactionId frozenxid;
+		MultiXactId	minmxid;
+	
+		frozenxid = relform1->relfrozenxid;
 		relform1->relfrozenxid = relform2->relfrozenxid;
-		relform2->relfrozenxid = swaptempxid;
+		relform2->relfrozenxid = frozenxid;
 
-		swaptempxid = relform1->relminmxid;
+		minmxid = relform1->relminmxid;
 		relform1->relminmxid = relform2->relminmxid;
-		relform2->relminmxid = swaptempxid;
+		relform2->relminmxid = minmxid;
 	}
 
 	/* swap size statistics too, since new rel has freshly-updated stats */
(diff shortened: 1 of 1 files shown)


lib/repack.c
```diff
@@ -1189,6 +1189,7 @@ swap_heap_or_index_files(Oid r1, Oid r2)
 	Form_pg_class relform1,
 				relform2;
 	Oid			swaptemp;
+	TransactionId swaptempxid;
```
MasahikoSawada, line 1192: `swaptempxid` can be declared inside of the `if (relform1->relkind != RELKIND_INDEX)` block.

lib/repack.c
```diff
@@ -1225,20 +1226,25 @@ swap_heap_or_index_files(Oid r1, Oid r2)
...
-		if (TransactionIdFollows(relform1->relfrozenxid,
-								 relform2->relfrozenxid))
-			relform1->relfrozenxid = relform2->relfrozenxid;
-		else
-			relform2->relfrozenxid = relform1->relfrozenxid;
+		swaptempxid = relform1->relfrozenxid;
+		relform1->relfrozenxid = relform2->relfrozenxid;
+		relform2->relfrozenxid = swaptempxid;
+
+		swaptempxid = relform1->relminmxid;
```
MasahikoSawada, line 1238: I think we should use a `MultiXactId` variable instead.

MasahikoSawada (commented):
I agree with overall of these changes, but I have some minor comments.

za-arthur:
Thank you @MasahikoSawada for the review! I pushed changes.

lib/repack.c
za-arthur: I used `TransactionId` mainly because `relminmxid` has that type, regardless of the comment in `pg_class.h`, stating that it is a `MultiXactId`.
I wonder why `pg_class.h` doesn't use `MultiXactId` for `relminmxid`.

MasahikoSawada: > I wonder why pg_class.h doesn't use MultiXactId for relminmxid.

I think MultiXactId doesn't have the corresponding SQL data type unlike TransactionId. Data types that can be used in system catalog are somewhat limited. It makes sense to use MultiXactId internally since we use it in C functions.

MasahikoSawada:
The changes look good to me. I'll merge it, barring any objections.

Issue #382, opened 2024-02-27
Passwords in .pgpass with spaces cause error Unable to set up worker conn #0: missing "=" 


jschaf:
I have a pgpass file that contains a password with spaces, like so:

```
my-host.com:5432:postgres:postgres:alpha bravo charlie
```

When running pg_repack, I see a warning in the logs that leaks the second word of the password, `bravo`

```
INFO: Dry run enabled, not executing repack
NOTICE: Setting up workers.conns
WARNING: Unable to set up worker conn #0: missing "=" after "bravo" in connection info string
INFO: repacking table "table1"
INFO: repacking table "table2"
...
```

It seems like the command still works, but reporting since pg_repack shouldn't leak partial passwords.

Taking a glance at the code, the problem seems to be that appending the password in `setup_workers` doesn't quote strings with whitespace at:

https://github.com/reorg/pg_repack/blob/3ecda64dbf3084b9071a44d893ffa3f1487c2193/bin/pgut/pgut-fe.c#L65

So the conn string parser sees something like:

```
host=my-host.com password=alpha bravo charlie
```

The parser attempts to interpret `bravo` as a key instead of part of the password string.

za-arthur:
@jschaf thank you for the report.
This should be fixed by the PR https://github.com/reorg/pg_repack/pull/384. Would you have a chance to check it?

PR #384, opened 2024-02-28
Replace `PQconnectdb()` by `PQconnectdbParams()`

za-arthur:
A value in a connection string should be quoted if it can contain spaces:
https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-KEYWORD-VALUE

Alternative approach is to use `PQconnectdbParams()` instead of `PQconnectdb()`. This should be more solid approach to pass password and other connection options, which doesn't require escaping value strings.

Issue: #382

### DIFF
--- a/bin/pgut/pgut-fe.c
+++ b/bin/pgut/pgut-fe.c
@@ -157,21 +157,9 @@ void disconnect_workers(void)
 void
 reconnect(int elevel)
 {
-	StringInfoData	buf;
 	char		   *new_password;
 
 	disconnect();
-	initStringInfo(&buf);
-	if (dbname && dbname[0])
-		appendStringInfo(&buf, "dbname=%s ", dbname);
-	if (host && host[0])
-		appendStringInfo(&buf, "host=%s ", host);
-	if (port && port[0])
-		appendStringInfo(&buf, "port=%s ", port);
-	if (username && username[0])
-		appendStringInfo(&buf, "user=%s ", username);
-	if (password && password[0])
-		appendStringInfo(&buf, "password='%s' ", password);
 
 	connection = pgut_connect(dbname, host, port, username, password, prompt_password, elevel);
 	conn2      = pgut_connect(dbname, host, port, username, password, prompt_password, elevel);
@@ -187,8 +175,6 @@ reconnect(int elevel)
 			password = pgut_strdup(new_password);
 		}
 	}
-
-	termStringInfo(&buf);
 }
 
 void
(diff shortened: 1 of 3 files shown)


bin/pgut/pgut-fe.c
```diff
@@ -62,7 +62,7 @@ setup_workers(int num_workers)
  		if (username && username[0])
  			appendStringInfo(&buf, "user=%s ", username);
  		if (password && password[0])
- 			appendStringInfo(&buf, "password=%s ", password);
+ 			appendStringInfo(&buf, "password='%s' ", password);
```
jschaf, line 65: That'll work for most passwords (and mine), but this approach won't work for passwords that contain single quotes.

https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-KEYWORD-VALUE

To make it bulletproof: when wrapping with single quotes, we'd need to scan the string and:
- Prefix any backslash with a backslash.
- Prefix any single quote with a backslash.

jschaf (commented):
This would fix my issue--left a suggestion for covering all cases.

bin/pgut/pgut-fe.c
za-arthur: That makes sense :+1:. I see couple of ways of resolving other cases:
- The simpler approach is to copy the function [appendConnStrVal()](https://github.com/postgres/postgres/blob/41f18f021a0882eccbeca62e2ed4b66c6b96e9c9/src/fe_utils/string_utils.c#L485-L521). `libpgfeutils` started sharing it only after Postgres 9.6.
- IMHO the proper approach would be to use [PQconnectdbParams()](https://github.com/postgres/postgres/blob/0028b55bc8e2fc80816094620d116e2cbbd8e5f3/src/interfaces/libpq/fe-connect.c#L668-L679), which doesn't require connection parameters values and which is used by most of the PostgreSQL client applications.

za-arthur: With the commit https://github.com/reorg/pg_repack/pull/384/commits/135df8b6a6a3b145439315b18539e583cf9fec5c I replaced call of `PQconnectdb()` by `PQconnectdbParams()`. This should be more solid approach to pass password and other connection options, which doesn't require escaping value strings.

jschaf: Much appreciated! I don't see anything glaringly wrong, but I have very little C experience.

andreasscherbaum (approved):
Looks good. Thanks for keeping older versions alive.

za-arthur:
Thanks for the approval. I'm merging the PR.

Issue #386, opened 2024-03-10
version 1.5.0 pg_repack the pg_repack command returns an ERROR: the request was not executed: ERROR: the tablespace ""XXXX"" does not exist


976Evill:
version 1.5.0 pg_repack the pg_repack command returns an ERROR: the request was not executed: ERROR: the tablespace ""XXXX"" does not exist

Steps for playback:
The name of the tablespace must start with a number.  The error example is below:

CREATE A TABLESPACE "2c_cfs" AT THE LOCATION "/2c_cfs" WITH (compression=true);
CREATE THE appdb TABLESPACE OF THE "2c_cfs" DATABASE;

\c appdb
create the pg_repack extension;

CREATE A workbook TABLE (
PRIMARY KEY book_id INTEGER,
title text,
integer prices
);

INSERT INTO books(book_id, name, price)
values
('101', 'Jobs', '2000'),
('102', ' Gita', '250'),
('103', ' Ramayana', '354'),
('104', ' Vedas', '268');

delete from books;

pg_repack -U postgres -d appdb -t books --tablespace "2c_cfs" -e --elevation=debug

LOG: (query) SELECT THE pid FROM pg_locks, WHERE locktype = 'relation' And granted = false And relation = 16788 And mode = 'AccessExclusiveLock' AND pid <> pg_backend_pid()
DEBUGGING: There is no competing DDL to cancel.
LOG: (request) SET LOCAL lock_timeout = 100
LOG: (request) LOCK THE public.books TABLE IN SHARED ACCESS MODE
LOG: (request) RESET lock_timeout
LOG: (query) SELECT repack.create_table($1, $2)
LOG: (parameter:0) = 16788
journal: (parameter:1) = "2c_cfs"
ERROR: query failed: ERROR: tablespace ""2c_cfs"" does not exist
CONTEXT: SQL statement "CREATE A TABLE repack.table_16788 WITH A TABLESPACE (oids = false) " ""2c_cfs""" HOW TO SELECT book_id, name,price ONLY FROM public.books WITHOUT DATA"
Function PL/pgSQL repack.create_table(oid,name) line 3 when EXECUTING
DETAILS: the request was: SELECT repack.create_table($1, $2)
DEBUGGING: There are no work items to disable.
LOG: (query) SET search_path TO pg_catalog, pg_temp, public
LOG: (query) SET search_path TO pg_catalog, pg_temp, public
LOG: (request) START ISOLATION LEVEL READING IS FIXED
[...]

PR #387, opened 2024-03-11
Issue 386: Do not call quote_ident() within repack_one_database()

za-arthur:
It isn't necessary to call `quote_ident()` within `repack_one_database()`. `tablespace_dest` is passed to `repack.create_table()`, which already calls `quote_ident()`.

`quote_ident()` within `repack_one_database()` can return quoted `tablespace_dest`, calling `quote_ident()` second time will make unexpected tablespace name with surrounding quotes.

Tests were added.

Issue: https://github.com/reorg/pg_repack/issues/386

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -34,14 +34,8 @@ jobs:
       - name: Put pg_repack on PATH
         run: echo "$PWD/bin" >> $GITHUB_PATH
 
-      - name: Create testts directory
-        run: sudo -u postgres mkdir /tmp/testts /tmp/1testts
-
-      - name: Create testts tablespace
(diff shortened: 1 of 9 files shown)


.github/workflows/regression.yml
```diff
@@ -35,11 +35,14 @@ jobs:
         run: echo "$PWD/bin" >> $GITHUB_PATH
 
       - name: Create testts directory
-        run: sudo -u postgres mkdir /tmp/testts
+        run: sudo -u postgres mkdir /tmp/testts /tmp/1testts
```
andreasscherbaum, line 38: While on it, can you add a few more "fancy" tablenames, like with spaces and other non-normal characters?
This is a quick way to see if and where this breaks.

andreasscherbaum (approved):
Looks good.

za-arthur:
Thanks for the approval. Merging.

PR #392, opened 2024-04-12
Add apply-count parameter

alex-richman-onesignal:
- Add a new apply-count command line parameter to allow tuning of the REDO apply_count variable.
- Check on startup that switch_threshold is not greater or equal to apply_count, which would cause REDO to stop after the first apply_log call (fwiw this bug also exists in 1.5.0, since switch_threshold parameter was added but was not checked against the default APPLY_COUNT of 1000).
- Tuning apply_count might be useful for improving REDO performance on tables with small tuples that receive lots of updates.
- Fixes https://github.com/reorg/pg_repack/issues/393

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -43,10 +43,10 @@ const char *PROGRAM_VERSION = "unknown";
 
 
 /*
- * APPLY_COUNT: Number of applied logs per transaction. Larger values
+ * APPLY_COUNT_DEFAULT: Number of applied logs per transaction. Larger values
  * could be faster, but will be long transactions in the REDO phase.
  */
-#define APPLY_COUNT		1000
+#define APPLY_COUNT_DEFAULT		1000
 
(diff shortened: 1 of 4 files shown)


magec (approved):

za-arthur:
Thanks! No objections from my side.

Could you add documentation and tests like here: https://github.com/reorg/pg_repack/pull/347/files#diff-1b3cd3a9c0b8b485f72c559854e6db0705069ce3e995fa25972825d9dcae359a.

andreasscherbaum:
Agree with Artur. A test would be nice to have. Otherwise it's a good addition.

alex-richman-onesignal:
Thanks both, added a few tests and a doc line.

za-arthur:
Thanks for the fix. I'm merging the PR.

za-arthur:
AFAIK we don't have a release schedule. We could make a release after accumulating few more fixes.

Issue #389, opened 2024-03-26
repacking partitioned table indexes failed with: Could not find index/cannot create index on partitioned table concurrently


ZhangHien:
I used pg_repack 1.5.0 to repack the indexes of parent table and its children on both PG11 and PG14, and everything worked as expected:
```sql
CREATE TABLE parent(val INTEGER PRIMARY KEY);
CREATE TABLE child_1(val INTEGER PRIMARY KEY) INHERITS(parent);
CREATE TABLE child_2(val INTEGER PRIMARY KEY) INHERITS(parent);

pg_repack --dbname=postgres --parent-table=parent --only-indexes
INFO: repacking indexes of "public.child_1"
INFO: repacking index "public.child_1_pkey"
INFO: repacking indexes of "public.child_2"
INFO: repacking index "public.child_2_pkey"
INFO: repacking indexes of "public.parent"
INFO: repacking index "public.parent_pkey"
```

However, when I used the same pg_repack command on a partitioned table, I encountered errors.
The definition of the partitioned table is as follows:
```sql
CREATE TABLE partitioned (val INT PRIMARY KEY) PARTITION BY RANGE (val);
CREATE TABLE partition_1 PARTITION OF partitioned FOR VALUES FROM (0) TO (1000);
CREATE TABLE partition_2 PARTITION OF partitioned FOR VALUES FROM (1000) TO (2000);
CREATE TABLE partition_default PARTITION OF partitioned DEFAULT;
```

On PG14, the **repack_index_swap function** could not find the new index index_16766 during rebuilding the index on the partitioned table because it only looks for indexes whose relkind is 'i', while the **index on partitioned table has relkind 'I'**, so the index could not be found.
```
pg_repack --dbname=postgres --parent-table=partitioned --only-indexes
INFO: repacking indexes of "public.partition_1"
INFO: repacking index "public.partition_1_pkey"
INFO: repacking indexes of "public.partition_2"
INFO: repacking index "public.partition_2_pkey"
INFO: repacking indexes of "public.partition_default"
INFO: repacking index "public.partition_default_pkey"
INFO: repacking indexes of "public.partitioned"
INFO: repacking index "public.partitioned_pkey"
ERROR: query failed: ERROR:  Could not find index 'index_16766', found 0 matches
DETAIL: query was: SELECT repack.repack_index_swap($1)
```

[...]

PR #398, opened 2024-05-16
Issue #389: Do not include a declaratively partitioned table with option `--only-indexes`

za-arthur:
When the option `--only-indexes` is used it doesn't make sense to repack indexes of a decoratively partitioned table since they don't store data. Therefore we can just ignore parent tables within the function `repack_all_indexes()`.

Issue: https://github.com/reorg/pg_repack/issues/389

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -2297,7 +2297,7 @@ repack_all_indexes(char *errbuf, size_t errsize)
 			res = execute_elevel("SELECT quote_ident(n.nspname) || '.' || quote_ident(c.relname)"
 								 " FROM pg_class c JOIN pg_namespace n on n.oid = c.relnamespace"
 								 " WHERE c.oid = ANY (repack.get_table_and_inheritors($1::regclass))"
-								 "   AND c.relkind != 'p'"
+								 "   AND c.relkind = 'r'"
 								 " ORDER BY n.nspname, c.relname",
 								 1, params, DEBUG2);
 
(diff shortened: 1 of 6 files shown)


andreasscherbaum:
Looks good, except tests for older versions are failing.

bin/pg_repack.c
```diff
@@ -2285,11 +2285,21 @@ repack_all_indexes(char *errbuf, size_t errsize)
...
+			 * the parent table. If the parent table is a declaratively
+			 * partitioned table then it isn't included into the result. It
+			 * doesn't make sense to repack indexes of a declaratively partitioned
+			 * table.
+			 */
 			res = execute_elevel("SELECT quote_ident(n.nspname) || '.' || quote_ident(c.relname)"
 								 " FROM pg_class c JOIN pg_namespace n on n.oid = c.relnamespace"
 								 " WHERE c.oid = ANY (repack.get_table_and_inheritors($1::regclass))"
-								 " ORDER BY n.nspname, c.relname", 1, params, DEBUG2);
+								 "   AND c.relkind != 'p'"
```
Melkij, line 2300: do we support anything other than relkind='r'?

za-arthur:
> Looks good, except tests for older versions are failing.

Yeah. Looks like Postgres 10 and older versions don't support some features. Postgres 10 supports declarative partitioning, but it doesn't support indexes on partitioned tables ("parents"). I guess I'd need to rewrite the test to be able to test `--only-indexes` on partitions on Postgres 10 too.

bin/pg_repack.c
za-arthur: Good point. I think I can change the condition to `c.relkind = 'r'`, because `--only-indexes` doesn't support other options anyway.

Melkij (approved):
Looks good.

za-arthur:
Merging the PR.

Issue #399, opened 2024-05-18
Silent data loss due to two vacuum running logically different tables using the same relfilenode.


MasahikoSawada:
Hi,

I've observed a silent data loss problem if (auto)vacuum, swapping tables by pg_repack, and INSERT are executed concurrently. The root cause is that while pg_repack is swapping the target table and its temp table, these two tables have different OID but use the same underlying file (i.e., use the same relfilenode), and two vacuum processes can concurrently process the same underlying file. I've confirmed this problem with PostgreSQL 15.4 and pg_repack 1.4.8. I think it impacts other PostgreSQL versions and pg_repack versions as well.

The reproducible step is:

# Preparation

```sql
create table test (id int primary key, a int);
insert into test values (1, 1), (2,2), (3,3);
```

These tuples are inserted at TID (0, 1), (0, 2), and (0, 3) on the `test table, respectively.

# Steps

1. Run pg_repack with debugger for the table A, and stop pg_repack after copying table data to its temp table by creating a breakpoint, for example at  [pg_repack.c:1537](https://github.com/reorg/pg_repack/blob/ver_1.4.8/bin/pg_repack.c#L1537).
2. Update a tuple of `test` table A, say `UPDATE test SET id = 999, a = 999 WHERE id = 1`. This change is logged to its log table.
3. Let pg_repack resume, its applies the log. And stop pg_repack after all logs are applied, for example at [pg_repack.c:L1598](https://github.com/reorg/pg_repack/blob/ver_1.4.8/bin/pg_repack.c#L1598). 
    -  In the temp table, the tuple at (0, 1) becomes garbage now.
    - Note that `conn2` already holds an AccessExclusiveLock on the `test` table A, but nobody has a lock on the temp table.
4. Run vacuum on the temp table and stop it before calling to [`lazy_vacuum_heap()`](https://github.com/postgres/postgres/blob/REL_15_4/src/backend/access/heap/vacuumlazy.c#L1264).
    - vacuum scans the temp table and collects TID (0,1) as dead tuple, but does not actually remove the itemid there yet. 
5. Let pg_repack resume, it swaps the `test` table and the temp table. And stop pg_repack right after swapping them, for example at [pg_repack.c:1617](https://github.com/reorg/pg_repack/blob/ver_1.4.8/bin/pg_repack.c#L1617).
    - Now both the table test and the temp table use the same relfilenode. And `conn2` already released AccessExclusiveLock on the `test` table.
[...]

PR #400, opened 2024-05-20
Fix possible two vacuums concurrently processing the same relfilenode.

MasahikoSawada:
Previously, while swapping the target table and its temp table, we used to acquire an AccessExclusiveLock on only the target table but not on its temp table. So the following scenario could happen:

1. Run (auto) vacuum on the temp table.
2. Swap the target table's relfile node and the temp table's relfilenode, and commit.
3. Run (auto) vacuum on the target table.

1 and 3 run concurrently on different tables but these tables use the same relfilenode. In the reported case, since concurrent changes (INSERT/UPDATE/DELETE) can also run concurrently, the second vacuum ended up removing newly added tuples.

To fix the issue, this commit changes the pg_repack command so that it acquires an AcessExclusiveLock also on the temp table before swapping these tables.

Issue: #399

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -911,6 +911,7 @@ repack_one_database(const char *orderby, char *errbuf, size_t errsize)
 		c++; // Skip schemaname
 		table.pkid = getoid(res, i, c++);
 		table.ckid = getoid(res, i, c++);
+		table.temp_oid = InvalidOid; /* filled after creating the temp table */
 
 		if (table.pkid == 0) {
 			ereport(WARNING,
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -189,6 +189,7 @@ typedef struct repack_table
 	Oid				target_tidx;	/* target: toast index OID */
 	Oid				pkid;			/* target: PK OID */
 	Oid				ckid;			/* target: CK OID */
+	Oid				temp_oid;		/* temp: OID */
 	const char	   *create_pktype;	/* CREATE TYPE pk */
 	const char	   *create_log;		/* CREATE TABLE log */
 	const char	   *create_trigger;	/* CREATE TRIGGER repack_trigger */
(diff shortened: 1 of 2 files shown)


Melkij:
Thanks! It looks like pg_reorg originally relied on [autovacuum=off](https://github.com/reorg/pg_repack/blame/ver_1.5.0/bin/pg_repack.c#L1528) , but that's really not enough (anti-wraparound autovacuum can still start). Explicit blocking is much more correct.
Will merge in a couple of days if there are no objections.

MasahikoSawada:
Thank you for looking at the fix.

 I guess we might want to check in `repack_swap()` if two relations are locked in AccessExclusiveMode for sure, for example by using `CheckRelationLockedByMe()`. What do you think?

Melkij:
sounds like good check. Although for now it should be conditional `#if PG_VERSION_NUM >= 120000`. [CheckRelationLockedByMe](https://github.com/postgres/postgres/commit/b04aeb0a053e7cf7faad89f7d47844d8ba0dc839) does not exist in all versions we still support.

za-arthur:
Thank you for the PR!

Would it make sense to assign `0` to `temp_oid` after initializing `repack_table`?
https://github.com/reorg/pg_repack/blob/f5aa3884a655fc0e8fa20ab25cbf91aba228fcdf/bin/pg_repack.c#L902

> I guess we might want to check in repack_swap() if two relations are locked in AccessExclusiveMode for sure, for example by using CheckRelationLockedByMe(). What do you think?

Indeed, it might be a good addition. I think instead `Assert` that check should raise a runtime error.

MasahikoSawada:
Thank you for reviewing the change! I've pushed the changes.

I use `LockHeldByMe()` instead of `CheckRelationLockedByMe()` since we don't open relations. These checks are available only on > PG12.

za-arthur (approved):
Thanks for the fix! Looks good to me.

MasahikoSawada:
Thank you for reviewing the change! I've merged it.

Issue #383, opened 2024-02-28
LOCK TABLE can only be used in transaction blocks


ZhangHien:
> I'm sorry that English is not my first language. I will try my best to describe this issue.

I am using pg_repack 1.4.8 to repack a partitioned table which has 100 partitions in PG 14.10:
`pg_repack -Utest_user --dbname=ddl_full0 -k --parent-table=part_range_2_level.t1`

Some partitions were repacked successfully at the beginning. But I got ERROR and WARNING when repacking the partition t1_p_0_p_2:
```
# ok
INFO: repacking table "part_range_2_level.t1_p_0_p_0"
INFO: repacking table "part_range_2_level.t1_p_0_p_1"
INFO: repacking table "part_range_2_level.t1_p_0_p_10"
NOTICE: Waiting for 1 transactions to finish. First PID: 28374

# ERROR and WARNING
INFO: repacking table "part_range_2_level.t1_p_0_p_2"
NOTICE: Canceled 1 unsafe queries
WARNING:  SET LOCAL can only be used in transaction blocks
WARNING: ERROR:  LOCK TABLE can only be used in transaction blocks

# Ignore the error above and repack other partitions
INFO: repacking table "part_range_2_level.t1_p_0_p_3"
NOTICE: Waiting for 6 transactions to finish. First PID: 10030218
INFO: repacking table "part_range_2_level.t1_p_0_p_4"
......
```
I find that pg_repack tried to get share lock on partition t1_p_0_p_2 for twice. It failed with timeout the first time and retried. The second time it failed with the error `LOCK TABLE can only be used in transaction blocks`:
```
# first lock request, fail with timeout
2024-02-23 11:24:50.837 UTC 30096 ddl_full0  55P03 0 ERROR:  canceling statement due to lock timeout
2024-02-23 11:24:50.837 UTC 30096 ddl_full0  55P03 0 STATEMENT:  LOCK TABLE part_range_2_level.t1_p_0_p_2 IN ACCESS SHARE MODE

# second lock request, fail with ERROR: LOCK TABLE can only be used in transaction blocks
2024-02-23 11:24:50.984 UTC 30096 ddl_full0  25P01 0 WARNING:  SET LOCAL can only be used in transaction blocks
2024-02-23 11:24:51.018 UTC 30096 ddl_full0  25P01 0 ERROR:  LOCK TABLE can only be used in transaction blocks
[...]

PR #401, opened 2024-05-20
Use savepoints when retrying to take AccessShareLock.

MasahikoSawada:
In lock_access_share() function, before acquiring an AccessShareLock on the target table, concurrent DDL commands are terminated unconditionally. However, there is a window between this termination and the actual "LOCK TABLE ... IN ACCESS SHARE MODE" operation, allowing other DDL commands to run on the target table. If the LOCK TABLE failed due to a lock conflict, the transaction was rolled back and retried to execute LOCK TABLE, but the LOCK TABLE command could not be executed again as the transaction had already closed.

This commit fixes the issue by using SAVEPOINT and ROLLBACK TO SAVEPOINT so that we can retry to take a lock in the same top transaction.

Issue #383

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -1753,7 +1753,8 @@ kill_ddl(PGconn *conn, Oid relid, bool terminate)
 /*
  * Try to acquire an ACCESS SHARE table lock, avoiding deadlocks and long
  * waits by killing off other sessions which may be stuck trying to obtain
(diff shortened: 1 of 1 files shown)


za-arthur (approved):
Thanks for the PR!. It looks reasonable to me.

MasahikoSawada:
Thank you for reviewing the change! Merged.

PR #409, opened 2024-07-12
Add doc in Chinese

yansheng836:
Add doc in Chinese

### DIFF
--- a/doc/pg_repack_zh.rst
+++ b/doc/pg_repack_zh.rst
@@ -1,0 +1,513 @@
+pg_repack -- 使用最小锁定重组 PostgreSQL 数据库中的表
+=====================================================
+
+.. contents::
+    :depth: 1
+    :backlinks: none
+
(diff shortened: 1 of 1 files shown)


andreasscherbaum:
@yansheng836 We've been discussing this patch. Is this a one-off translation, or do you plan to keep it updated if the need arrives to add further translations?

Also is there someone else who can contribute to this translation? We've had translation added in the past, and then abandoned.

yansheng836:
Yes, I will always keep updating.

dvarrazzo:
I would like to thank @yansheng836 for the proposal, but the translation in Chinese of the readme, but:

- this document is not published anywhere (the website code is not updated to expose it);
- we don't have a workflow for translations;
- needing to translate the docs makes releasing every feature more complex;
- if the translation doesn't happen in time, an user relying on it will just miss it, making a disservice to them and not reflecting well on us;
- I don't want to make wild generalization on the demographics of who uses a project, but I tend to believe that an user advanced enough as to needing to run pg_repack knows enough English to read our readme;
- none of us knows Chinese, so, as far as I know, the translation could be a list of swearing.

So I don't think we should merge this MR.

Note that the Japanese translation has similar shortcomings (with the difference that the original pg_reorg was written by Japanese people, so it made sense for them). If the latest versions have not been updated I propose to drop it too.

yansheng836:
Ok, I see.

Issue #426, opened 2024-10-16
Segmentation fault


JennyJennyChen:
session A:
pgbench -i -s 1000 pgbenchdb 
pg_repack -d pgbenchdb   -t pgbench_accounts -k

Start session B connect DB during the repack process 
session B:
alter system set default_transaction_read_only=on;
SELECT pg_reload_conf();

session A repack enters an infinite loop, error stack burst, coredump
![image](https://github.com/user-attachments/assets/df55468c-d2ee-409c-bd64-1b595c4ee64d)

andreasscherbaum:
> alter system set default_transaction_read_only=on;

Why would you do that?

JennyJennyChen:
> > alter system set default_transaction_read_only=on;
> 
> Why would you do that?

Because when the disk capacity is insufficient, the automatic protection is triggered, all write transactions are interrupted and write transactions are eliminated. Anyway, no matter under any circumstances and any settings, coredump should not occur.

za-arthur:
The issue was fixed by the PRs: https://github.com/reorg/pg_repack/pull/427, https://github.com/reorg/pg_repack/pull/431

PR #427, opened 2024-10-16
Fixed segmentation fault

JennyJennyChen:
Fixed the bug that segmentation fault occurs when default_transaction_read_only is turned on during repacking. 
bug: https://github.com/reorg/pg_repack/issues/426


za-arthur:
Thanks for the PR! The commit was merged into `master` together with the PR https://github.com/reorg/pg_repack/pull/431.

Issue #419, opened 2024-09-01
Endless recursive call of repack_cleanup_callback() when lock failed


JohnHien:
Hi all, I used pg_repack 1.5.0 to repack a normal table and I encountered endless recursive call of repack_cleanup_callback() if I failed to lock the table, finally I got a coredump file.

Here are my SQL and shell commands. I'm sorry that you cannot reproduce it in community version of PG and pg_repack because I used it in a cloud database Alibaba Cloud PolarDB-PG. In this db there is a kind of quasi-superuser which has more privileges than normal user but less privileges than real superuser, whose name is POLAR_SUPERUSER. If the user doesn't have the privilege of the table, we can reproduce the error case because we cannot lock the table.

Create normal table and extension with real superuser:
```sql
postgres=# create table test(a int primary key);
CREATE TABLE
postgres=# create extension pg_repack ;
CREATE EXTENSION
```

Create a POLAR_SUPERUSER, grant the privileges of schema repack to it. But it doesn't have the privilege of the normal table.
```sql
postgres=# create user test_user with polar_superuser;
CREATE ROLE
postgres=# grant all on schema repack to test_user ;
GRANT
postgres=# grant all on all tables in schema repack to test_user ;
GRANT
```

Then repack the table with the user created in last step and we can get the error messages.
```sh
pg_repack --dbname=postgres --user test_user -k --table=test
```

The error messages and coredump stack are as follows.
```
INFO: repacking table "public.test"
ERROR:  permission denied for table test
ERROR: query failed: ERROR:  current transaction is aborted, commands ignored until end of transaction block
DETAIL: query was: RESET lock_timeout
ERROR:  permission denied for table test
ERROR: query failed: ERROR:  current transaction is aborted, commands ignored until end of transaction block
……
……
ERROR: query failed: ERROR:  current transaction is aborted, commands ignored until end of transaction block
DETAIL: query was: RESET lock_timeout
ERROR:  permission denied for table test
ERROR: query failed: ERROR:  current transaction is aborted, commands ignored until end of transaction block
DETAIL: query was: RESET lock_timeout
ERROR:  permission denied for table test
ERROR: query failed: ERROR:  current transaction is aborted, commands ignored until end of transaction block
DETAIL: query was: RESET lock_timeout
Segmentation fault (core dumped)
```

```
(gdb) bt
#0  0x00007f1cddc74e46 in pg_snprintf (str=<error reading variable: Cannot access memory at address 0x7ffd1840df78>,
[...]

Issue #223, opened 2020-01-09
--no-superuser-check does not work


bradnicholson:
The `--no-superuser-check` does not work.  When passed in as an option, it fails with:

```
pg_repack -k  -U admin -d test  -t foo
INFO: repacking table "public.foo"
ERROR: query failed: ERROR:  must be superuser to use repack_apply function
DETAIL: query was: SELECT repack.repack_apply($1, $2, $3, $4, $5, $6)
ERROR: query failed: ERROR:  must be superuser to use repack_drop function
DETAIL: query was: SELECT repack.repack_drop($1, $2)
ERROR: query failed: ERROR:  must be superuser to use repack_drop function
DETAIL: query was: SELECT repack.repack_drop($1, $2)
ERROR: query failed: ERROR:  must be superuser to use repack_drop function
DETAIL: query was: SELECT repack.repack_drop($1, $2)
ERROR: query failed: ERROR:  must be superuser to use repack_drop function
DETAIL: query was: SELECT repack.repack_drop($1, $2)
```

Looking at the code - its failing here https://github.com/reorg/pg_repack/blob/master/lib/repack.c#L111

This always evaluates to the actual superuser status of the user.

I don't see `superuser()` defined anywhere in pg_repack - so I presume that is coming in from Postgres includes.

Initial discussion in #201

dvarrazzo:
Thank you for looking into this. Can you provide a patch please?

schmiddy:
IIRC the `--no-superuser-check` was added so that pg_repack could work on AWS Postgres RDS. The permissions in RDS are a little different from vanilla Postgres - users don't have available a full-fledged superuser role, rather they can get a role that belongs to the `rds_superuser` group. It's possible Postgres RDS has a slightly different implementation of that `superuser()` function underneath the hood.

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.PostgreSQL.CommonDBATasks.html

I'm not opposed to a change/fix here, but it would be nice if we could avoid breaking functionality for users on RDS.

dvarrazzo:
We might event want to keep it simple and wontfix this, documenting that `--no-superuser-check` is for aws only

bradnicholson:
The binary correctly checks and blocks non-super user access, and correctly respects the `-k` flag.

Is the secondary check in the backend code actually needed?  Sure, it prevents non-superusers from calling the functions directly in the DB, but the binary (intends) to add that functionality anyway.

[...]

PR #431, opened 2024-10-30
Allow to run pg_repack by non-superuser

za-arthur:
The option `--no-superuser-check` allows to by-pass the check if the user is a superuser. That was done for users which run pg_repack on Amazon, where users cannot run it as a superuser.
The problem is that the option `--no-superuser-check` works only on the CLI level, skipping the check only by the `pg_repack` client. But there are also checks done by the extension functions exported to SQL.

The PR removes the check that the user is a superuser from functions `repack_trigger()` and `repack_apply()`. That check is redundant since queries executed by that functions can be executed by a user manually. Moreover `repack_trigger()` is a `SECURITY DEFINER` function, which means that it is executed with superuser privileges (`pg_repack` extension can be created only by superuser).

The PR changes privilege check in functions `repack_swap()`, `repack_drop()` and `repack_index_swap()`. Now that functions can be run by an owner of a table. That check is necessary since `_swap` functions swap relfilenodes on `pg_class` system catalog table. `repack_drop()` acquires ACCESS EXCLUSIVE lock and therefore it also requires privilege check.

Additionally I cherry-picked the commit 326b6e1 from the PR https://github.com/reorg/pg_repack/pull/427. Otherwise `pg_repack` will fall with segmentation fault in case of lack of permissions.

Relevant issues:
- https://github.com/reorg/pg_repack/issues/223
- https://github.com/reorg/pg_repack/issues/419
- https://github.com/reorg/pg_repack/issues/426

### DIFF
--- a/META.json
+++ b/META.json
@@ -2,7 +2,7 @@
    "name": "pg_repack",
    "abstract": "PostgreSQL module for data reorganization",
    "description": "Reorganize tables in PostgreSQL databases with minimal locks",
-   "version": "1.5.1",
+   "version": "1.5.2",
    "maintainer": [
        "Beena Emerson <memissemerson@gmail.com>",
        "Josh Kupershmidt <schmiddy@gmail.com>",
@@ -17,7 +17,7 @@
    "provides": {
       "pg_repack": {
          "file": "lib/pg_repack.sql",
-         "version": "1.5.1",
+         "version": "1.5.2",
          "abstract": "Reorganize tables in PostgreSQL databases with minimal locks"
       }
    },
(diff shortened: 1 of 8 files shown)


andreasscherbaum:
Ah, nice, comes with additional tests. I like that.

za-arthur:
I'd like to merge the PR this week if no objections.

za-arthur:
@ololobus new version was released https://github.com/reorg/pg_repack/tree/ver_1.5.2. Although the website wasn't updated yet.

ololobus:
@za-arthur thanks! The git tag is what I actually needed

Issue #422, opened 2024-09-25
Making --error-on-invalid-index the default behavior


zubeyr94:
Hello, currently the default behavior in pg_repack is to repack tables even if tables have invalid indexes on them. Users can force pg_repack to error when there are invalid indexes on tables before they are repacked by providing `-error-on-invalid-index` parameter on pg_repack 1.5.0 and later versions.

However, repacking tables with invalid indexes would cause the index to be corrupted, because repack will skip repacking the invalid index and old heap tuple id in the index now be pointing to incorrect tuples on the relation. As a safety feature it would make sense to make `--error-on-invalid-index` the default behavior and there could be an option like `--warn-on-invalid-index` or `--no-error-on-invalid-index` for users who want to keep the old behavior.

MasahikoSawada:
> However, repacking tables with invalid indexes would cause the index to be corrupted. As a safety feature it would make sense to make --error-on-invalid-index the default behavior

What do exactly mean by safety feature? It's right that invalid indexes could corrupted after repacking a table, but invalid indexes are not used unless they are reindexed.

zubeyr94:
Postgres continues to update indexes if indisready set to true, even if index is invalid, ie indisvalid is false. On a corrupted index, checks made during index insertion could cause transactions to rollback, eg [this check could fail](https://github.com/postgres/postgres/blob/7f7474a8e4002ac9fd4979cc7b16b50b70b70c28/src/backend/access/nbtree/nbtinsert.c#L1184). Until users realize this and drop/reindex the indexes they won't be able to unblock the transaction doing the insertion. Instead of taking the hit while running regular workload, it should be preferable to take the hit during maintenance work, ie while running pg_repack and make users aware that they need to reindex or drop before completing the repack.

MasahikoSawada:
I think that transaction rollbacks caused by inserting indexes tuples to invalid indexes can happen immediately since indexes becoming invalid, regardless of pg_repack. That is, if we make the `--error-on-invalid-index` behavior default, we will enforce users to rebuild or drop invalid indexes before pg_repack by default. I guess the current behavior is more flexible 
in terms of the timing of rebuilding or dropping invalid indexes.

Also, I'm concerned that this change would break the compatibility of tools that use `--error-on-invalind-index` option.

Another idea would be to have pg_repack use command options via an environment variable, say `PG_REPACK_OPTIONS`, and merge the options to the specified command options.

zubeyr94:
[...]

PR #433, opened 2024-11-12
Issue #422: Make --error-on-invalid-index the default behavior

zubeyr94:
Skipping repacking the invalid indexes could lead to corrupting the invalid indexes. In some rare cases, this might cause inserts to tables to fail as changes to a table continuously be applied to invalid indexes on it and updates to corrupted invalid indexes potentially might fail by the checks done during insertion. Therefore, by default do not repack a table with invalid indexes. Repack only user explicitly specifies --no-error-on-invalid-index.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -258,7 +258,9 @@ static unsigned int		temp_obj_num = 0; /* temporary objects counter */
 static bool				no_kill_backend = false; /* abandon when timed-out */
 static bool				no_superuser_check = false;
 static SimpleStringList	exclude_extension_list = {NULL, NULL}; /* don't repack tables of these extensions */
-static bool 			error_on_invalid_index = false; /* don't repack when invalid index is found */
+static bool 			no_error_on_invalid_index = false; /* repack even though invalid index is found */
+static bool 			error_on_invalid_index = false; /* don't repack when invalid index is found,
+								 * deprecated, this the default behavior now */
 static int				apply_count = APPLY_COUNT_DEFAULT;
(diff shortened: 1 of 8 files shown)


bin/pg_repack.c
```diff
@@ -1322,7 +1325,11 @@ repack_one_table(repack_table *table, const char *orderby)
 	{
 		const char *indexdef;
 		indexdef = getstr(indexres, j, 0);
-		if (error_on_invalid_index) {
+
+		/* Enforce --error-on-invalid-index in all cases unless
+		 * --no_error_on_invalid_index passed by itself
+		 */
+		if (error_on_invalid_index || !no_error_on_invalid_index) {
```
za-arthur, line 1332: Thanks for the PR! What if we completely ignore `error_on_invalid_index` here and make it no-op as it was suggested previously https://github.com/reorg/pg_repack/issues/422#issuecomment-2402913236, but keep it as a valid option and document it as a no-op in the documentation? WDYT?

andreasscherbaum: Needs to stay as an option anyway. Can be marked as deprecated, and in 2-3 years can be removed.

zubeyr94: Yeah, I think this makes sense. Released an updated version, which makes --error-on-invalid-index no-op and marked it as deprecated in helper and docs.

za-arthur: Thanks! I'd merge the PR if no objections.

za-arthur:
@zubeyr94 I merged the PR. Thanks!

PR #434, opened 2024-12-12
Release of pg_repack 1.5.2

za-arthur:
Looks like we are ready to make new release and there are no urgent issues which stops it.
Just add release notes. The version was already bumped within META.json before.


Melkij (approved):
look good

andreasscherbaum (approved):

Issue #437, opened 2025-01-20
Clear repack index artefacts when dropping the extension


flipsed:
We recently started using `--only-indexes` and started seeing messages like this:
```
WARNING: Cannot create index "public"."index_613963323", already exists
DETAIL: An invalid index may have been left behind by a previous pg_repack on the table which was interrupted. Please use DROP INDEX "public"."index_613963323" to remove this index and try again.
```

I can see that the [extension drop](https://github.com/reorg/pg_repack/blob/9f36c65bd57ca1b228025687843758556b56df8e/lib/repack.c#L1053) clears everything else but the indexes. We tend to do 
```
drop extension pg_repack cascade; create extension pg_repack;
```
if there are leftover repack artefacts but it hasn't been clearing the indexes

za-arthur:
@flipsed thank you for the report! I created a PR https://github.com/reorg/pg_repack/pull/440 to fix the issue.

za-arthur:
I merged the PR #440 and I'm closing this issue.

PR #440, opened 2025-01-26
Issue #437: Cleanup temporary indexes on error (with --only-indexes)

za-arthur:
When using `--only-indexes` `pg_repack` uses different approach to repack indexes, which is running CREATE INDEX CONCURRENTLY command to create a temporary index and swapping the target index and the temporary index afterwards.
That approach didn't use any cleanup callbacks and therefore in case of an error temporary indexes wouldn't be cleaned up.

The commit adds the callback `repack_cleanup_index`.

Issue: #437


za-arthur:
@andreasscherbaum @Melkij I'd merge this PR if no objections. There are no new tests since it will be necessary somehow to throw an error in the middle of repacking of an index.

PR #445, opened 2025-02-08
Lock down search_path of functions

svenklemm:
Functions referencing objects without schema are susceptible to accessing unintented objects depending on the search_path they are executed with. This patch adds a search_path to all the SQL and PLPGSQL functions to ensure the intended objects are accessed.

### DIFF
--- a/lib/pg_repack.sql.in
+++ b/lib/pg_repack.sql.in
@@ -22,7 +22,7 @@ CREATE FUNCTION repack.oid2text(oid) RETURNS text AS
 $$
 	SELECT textin(regclassout($1));
 $$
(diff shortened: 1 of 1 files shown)


za-arthur:
@svenklemm thank you for the contribution!

Currently `pg_repack` sets hardcoded `search_path` on every connection:
https://github.com/reorg/pg_repack/blob/306b0d4f7f86e807498ac00baec89ecd33411398/bin/pgut/pgut.c#L550

Does the PR improve handling of issue of calling unintented objects? BTW there are still views `repack.primary_keys` and `repack.tables`, which can call unintended objects. AFAIK it isn't possible to set `search_path` during `CREATE VIEW`, but because `pg_repack` already sets hardcoded `search_path` all views and functions calls should be handled.

PR #451, opened 2025-04-21
Grant extension-owner access to tables, functions in schema repack.

harinath001:
## Context

If PostgreSQL admin allows `pg_repack` to be trusted extension (y not granting superuser access to the users), then users should be ale to install and run pg_repack on their databases without the intervention of superusers/DBAs.

But today, when we make `pg_repack` as a trusted extension, superuser has to grant these privileges to the extension owners.

```
grant all on schema repack to non_superuser;
grant all privileges on all functions in schema repack to non_superuser;
grant all privileges on all tables in schema repack to non_superuser;
```

otherwise we get this error
```
ERROR: pg_repack failed with error: ERROR:  permission denied for schema repack
LINE 1: select repack.version(), repack.version_sql()
```

this is because when the extension is trusted, the schema is owned by install superuser

```
test_db=> \c test_db nonsuper
You are now connected to database "test_db" as user "nonsuper".
test_db=> create extension pg_repack;
CREATE EXTENSION
test_db=> \dn
      List of schemas
  Name  |       Owner       
--------+-------------------
 public | pg_database_owner
 repack | harinath
(2 rows)

test_db=> \du
                             List of roles
 Role name |                         Attributes                         
-----------+------------------------------------------------------------
 harinath  | Superuser, Create role, Create DB, Replication, Bypass RLS
 nonsuper  | Create DB

test_db=> 
```

Note: This patch should not affect the behavior if the extension is not trusted.


ardentperf:
can two different non-superusers both install separate copies of this extension at the same time, if they both need to repack a table?

Issue #420, opened 2024-09-01
ERROR in repack.repack_swap() if CIC creates a new index on the original table after rebuild_indexes() on new table


zhanghien:
In repack_one_table(), the rebuild_indexes() function is called to rebuild the indexes on the new table and the CREATE INDEX commands are from the index definition on the original table.
What if a new index is created on the original table after the rebuild_indexes() function called? I think this new index on the original table will not be created in the new table.

To prove it, I modified the code in repack_one_table() to sleep for 20s after calling rebuild_indexes(). When the pg_repack client was sleeping, I attached to it with gdb, so the client paused and I had time to create the new index.
I executed the CIC command and a new index test_a_idx was created on the original table. This is possible because only an ACCESS SHARE lock is held on the table now.
```sql
postgres=# create index concurrently on test(a);
CREATE INDEX
postgres=# \d+ test
                                          Table "public.test"
 Column |  Type   | Collation | Nullable | Default | Storage | Compression | Stats target | Description
--------+---------+-----------+----------+---------+---------+-------------+--------------+-------------
 a      | integer |           | not null |         | plain   |             |              |
Indexes:
    "test_pkey" PRIMARY KEY, btree (a)
    "test_a_idx" btree (a)
Triggers firing always:
    repack_trigger AFTER INSERT OR DELETE OR UPDATE ON test FOR EACH ROW EXECUTE FUNCTION repack.repack_trigger('a')
Access method: heap
```

Then I executed 'c' to continue in gdb and I got the error in repack_swap():
```
ERROR: query failed: ERROR:  relation "repack.index_16577" does not exist
CONTEXT:  SQL statement "SELECT X.oid, Y.oid  FROM pg_catalog.pg_index I,       pg_catalog.pg_class X,       pg_catalog.pg_class Y WHERE I.indrelid = $1   AND I.indexrelid = X.oid   AND I.indisvalid   AND Y.oid = ('repack.index_' || X.oid)::regclass"
DETAIL: query was: SELECT repack.repack_swap($1)
```

repack_swap() assumes that all the indexes on the original table are already created on the new table with name index_<oid>, but in this case the new index test_a_idx with oid 16577 was not created on the new table because it was created after rebuild_indexes().
```
postgres=# select relname from pg_class where oid = 16577;
  relname
------------
 test_a_idx
(1 row)
```

I know this is a corner case and I am not sure whether we should fix it.

za-arthur:
[...]

PR #452, opened 2025-04-21
Issue #420: Acquire SHARE UPDATE EXCLUSIVE by conn2

za-arthur:
Currently conn2 acquires `ACCESS SHARE` lock which allows to run concurrent DDL statements, like `CREATE INDEX CONCURRENTLY`.

Fix it by acquiring `SHARE UPDATE EXCLUSIVE` lock. This will also lock concurrent `VACUUM`, `ANALYZE` queries, locking them looks safer option anyway.

Thoughts?

Issue #420

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.o
 *.so
 *.bc
+*.dylib
 regress/regression.diffs
 regress/regression.out
 regress/results/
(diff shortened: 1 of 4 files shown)


andreasscherbaum:
Makes sense.

Issue #457, opened 2025-06-11
Subtransaction cache overflow with --no-kill-backend --wait-timeout=10800


asvrvs:
[pg_repack_assignment_table_log.txt](https://github.com/user-attachments/files/20689178/pg_repack_assignment_table_log.txt)

### Summary

When applying `pg_repack` in safe mode (`--no-kill-backend`) with a long wait timeout (specifically, 10800 sec) PostgreSQL cluster could degrade with so-called subtransactions (or [subtransactions cache overflow](https://www.rockdata.net/blog/subtransaction-cache-overflow/)) problem.

The (supposed) mechanics of the problem:

1. after applying CDC-log to a temporary table for the first time (and before the `repack_swap`) `pg_repack` tries to escalate the lock to AccessExclusive level;
2. to do this, it creates a SAVEPOINT (which opens a subtransaction):

```
LOG: (query) SAVEPOINT repack_sp1
LOG: (query) SET LOCAL statement_timeout = 100
LOG: (query) LOCK TABLE public.assignment IN ACCESS EXCLUSIVE MODE
```

3. if the load on the table is high, this (usually) doesn't work and subtransaction rolls back:

```
LOG: (query) ROLLBACK TO SAVEPOINT repack_sp1
```
4. and if the wait timeout is quite long (like, `--wait-timeout=10800`) SAVEPOINT … LOCK TABLE … ROLLBACK TO repeats N times (in our case – more than 3500);
5. crucial moment (IMHO) is how PostgreSQL [implements](https://www.postgresql.org/docs/current/sql-savepoint.html) SAVEPOINT: _SQL requires a savepoint to be destroyed automatically when another savepoint with the same name is established. In PostgreSQL, the old savepoint is kept, though only the more recent one will be used when rolling back or releasing._
6. when rolling back to a savepoint, PostgreSQL [doesn't release](https://www.postgresql.org/docs/current/sql-rollback-to.html) (delete) it: _The savepoint remains valid and can be rolled back to again later, if needed._
7. so, with multiple SAVEPOINT … LOCK TABLE … ROLLBACK TO PostgreSQL essentially creates multiple subtransactions with the same name (`repack_sp1`);
[...]

PR #460, opened 2025-07-06
Issue #457: Move SAVEPOINT outside the retry loop and call RELEASE SAVEPOINT

za-arthur:
`lock_access_share()` and `lock_exclusive()` doesn't call RELEASE SAVEPOINT on every retry which pollutes the cache and degrades the performance. This commit calls RELEASE SAVEPOINT when created savepoint is not necessary anymore. Additionally it moves SAVEPOINT outside of the retry loop since it isn't necessary to create new savepoint on every retry at all.

`lock_exclusive()` also moves start of transaction outside of the retry loop and creates a savepoint regardless of the value `start_xact`.

Now Github workflow also installs `libnuma-dev` for PostgreSQL 18 since it has new option `--with-libnuma`:
https://www.postgresql.org/docs/18/release-18.html
> Add configure option --with-libnuma to enable NUMA awareness (Jakub Wartak, Bertrand Drouvot)

Issue #457

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Install additional build-dependencies for 18+
         if: ${{ matrix.pg >= 18 }}
-        run: apt-get install -y libcurl4-openssl-dev
+        run: apt-get install -y libcurl4-openssl-dev libnuma-dev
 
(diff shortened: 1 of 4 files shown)


andreasscherbaum (approved):
looks good.

PR #480, opened 2026-02-27
Fix NOTICE suppression for pg_regress with PGAPPNAME

bayandin:
This PR fixes flaky regression tests caused by unsuppressed NOTICE messages
leaking into `make installcheck` output.

The NOTICE about waiting for transactions is meant to be suppressed when
running under pg_regress, but since PostgreSQL 10
(postgres/postgres@a4327296df7), pg_regress sets `PGAPPNAME` to
`pg_regress/<testname>` instead of just `pg_regress`. The existing
`strcmp` check only matched the exact string `"pg_regress"`, so the
suppression silently stopped working on PG 10+.

When a concurrent transaction happens to be present during the test run,
the NOTICE leaks into output and causes a spurious diff like:

```diff
@@ -380,6 +380,7 @@
 INFO: repacking table "public.child_a_2"
 INFO: repacking table "public.parent_a"
 INFO: repacking table "public.partition_a_1"
+NOTICE: Waiting for 1 transactions to finish. First PID: 572
 INFO: repacking table "public.partition_a_2"
```

- Match both `"pg_regress/<testname>"` (PG 10+) and bare `"pg_regress"` (PG <10) PGAPPNAME formats
- Update related comments to document both PGAPPNAME formats

Debugged and drafted the fix with assistance from Cursor IDE and Claude
claude-4.6-opus (Anthropic).

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -71,7 +71,7 @@ const char *PROGRAM_VERSION = "unknown";
  * based on pg_backend_pid() and application_name), but that shouldn't hurt
  * anything. Also, the test of application_name is not bulletproof -- for
  * instance, the application name when running installcheck will be
- * pg_regress.
+ * pg_regress/<testname>, or just pg_regress on PostgreSQL <10.
  */
 #define SQL_XID_SNAPSHOT_90200 \
 	"SELECT coalesce(array_agg(l.virtualtransaction), '{}') " \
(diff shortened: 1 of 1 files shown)


za-arthur:
@bayandin thank you for the fix.
It looks good to me. Although I'm curious do you reproduce this reliably? Locally I cannot reproduce it when I run `make installcheck`.

PR #481, opened 2026-03-17
pg_repack is aborted with error if a publication FOR ALL TABLES exist

Melkij:
pg_repack is aborted with error if a publication FOR ALL TABLES exists and apply_log requires deleting or updating a row in a temporary table.

Reproduction:
```
create table foo(id int primary key, val text);
insert into foo values (1, 'hi!');
create publication foo for all tables ;
# start pg_repack with break at rebuild_indexes
update foo set val = 'h2' where id = 1;
# continue pg_repack process and see error like this:

ERROR: query failed: ERROR:  cannot update table "table_16388" because it does not have a replica identity and publishes updates
HINT:  To enable updating the table, set REPLICA IDENTITY using ALTER TABLE.
CONTEXT:  SQL statement "UPDATE repack.table_16388 SET (id, val) = ($2.id, $2.val) WHERE (id) = ($1.id)"
DETAIL: query was: SELECT repack.repack_apply($1, $2, $3, $4, $5, $6)
```

It's not that difficult to format `ALTER TABLE repack.table_' || R.oid || ' REPLICA IDENTITY USING INDEX index_' || PK.indexrelid` in repack.tables, but for this to work, I also need to set NOT NULL on columns, which is less trivial. Furthermore, repack might break subscribers of this publication.

Given the lack of previous bug reports (this setup never worked since introduction of publications in pg10), I suggest adding a check for publications FOR ALL TABLES and simply refusing to work in such a database by default.

### DIFF
--- a/bin/pg_repack.c
+++ b/bin/pg_repack.c
@@ -258,6 +258,7 @@ static unsigned int		temp_obj_num = 0; /* temporary objects counter */
 static bool				no_kill_backend = false; /* abandon when timed-out */
 static bool				no_superuser_check = false;
 static SimpleStringList	exclude_extension_list = {NULL, NULL}; /* don't repack tables of these extensions */
+static bool 			no_error_on_publication = false; /* repack even though publication FOR ALL TABLES is found */
 static bool 			no_error_on_invalid_index = false; /* repack even though invalid index is found */
 static bool 			error_on_invalid_index = false; /* don't repack when invalid index is found,
 								 * deprecated, this the default behavior now */
@@ -292,6 +293,7 @@ static pgut_option options[] =
 	{ 'b', 'D', "no-kill-backend", &no_kill_backend },
 	{ 'b', 'k', "no-superuser-check", &no_superuser_check },
 	{ 'l', 'C', "exclude-extension", &exclude_extension_list },
+	{ 'b', 5, "no-error-on-publicationn", &no_error_on_publication },
 	{ 'b', 4, "no-error-on-invalid-index", &no_error_on_invalid_index },
 	{ 'b', 3, "error-on-invalid-index", &error_on_invalid_index },
 	{ 'i', 2, "apply-count", &apply_count },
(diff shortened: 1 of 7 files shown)


bin/pg_repack.c
```diff
@@ -559,6 +561,42 @@ preliminary_checks(char *errbuf, size_t errsize){
...
+            }
+            /* no publications, continue to work */
+        } else {
+            if (errbuf)
+                    snprintf(errbuf, errsize, "%s", PQerrorMessage(connection));
+            goto cleanup;
+        }
+        CLEARPGRES(res);
+    }
+
```
za-arthur: I suggest a minor change of formatting, mainly changing white spaces to tabs (which is used by most of the code):
```suggestion
	/*
	 * repack will be aborted with error like
	 *
	 * ERROR:  cannot update table "table_16388" because it does not
	 * have a replica identity and publishes updates
	 *
	 * if a publication FOR ALL TABLES exists and apply_log requires
	 * deleting or updating a row in a temp table. To make this works,
	 * we need to set NOT NULL on columns of primary key and command
	 *
	 * ALTER TABLE repack.table_' || R.oid || ' REPLICA IDENTITY USING INDEX index_' || PK.indexrelid
	 *
	 * In addition, repack might break subscribers of this publication.
	 * For now, we simply refuse to continue.
	 */
	if (PQserverVersion(connection) >= 100000 && !no_error_on_publication)
	{
		res = execute_elevel("SELECT pubname FROM pg_publication WHERE puballtables LIMIT 1",
			0, NULL, DEBUG2);
		if (PQresultStatus(res) == PGRES_TUPLES_OK)
		{
			if (PQntuples(res) > 0)
			{
				const char *pubname;
				pubname = getstr(res, 0, 0);
				snprintf(errbuf, errsize, "publication \"%s\" FOR ALL TABLES found: won't be able to apply concurrent UPDATE and DELETE", pubname);
				goto cleanup;
			}
			/* no publications, continue to work */
		} else {
			if (errbuf)
				snprintf(errbuf, errsize, "%s", PQerrorMessage(connection));
			goto cleanup;
		}
		CLEARPGRES(res);
	}
```

za-arthur (commented):
Makes sense. Thank you! Could you add new option to the documentation as well?
I have minor formatting suggestion below as well.

Melkij:
Thanks, I made the changes

Melkij:
I was in a hurry, I'll add a little more now...

Melkij:
Sorry for the noise, that's correct. _I'll go get some coffee..._

za-arthur (approved):
Thank you for the fixes. LGTM.

